PR #6, opened 2022-06-03
Leman rebased on master

nop33:


polarker:
Thanks for the rebasing, but I use merge instead. I don't think rebasing is a good practice and it does not scale in the long term. Please take a look at the git log of `web3.js`.

PR #19, opened 2022-08-09
Update schema

polarker:


src/api/api-alephium.ts
```diff
@@ -429,12 +429,14 @@ export interface CompileContractResult {
...
   functions: FunctionSig[]
   events: EventSig[]
+  warnings: string[]
 }
 
 export interface CompileScriptResult {
   bytecodeTemplate: string
   fields: FieldsSig
   functions: FunctionSig[]
+  warnings: string[]
```
Lbqds, line 439: Do we need to print warnings when compiling scripts and contracts?

Lbqds (approved):
LGTM, only one minor comment.

h0ngcha0 (approved):
LGTM :+1:

src/api/api-alephium.ts
polarker: I forgot to handle warning. I will create another PR. I am not sure if we should print the warnings or we should save it to a file for now.

PR #22, opened 2022-08-19
Fix unused variables errors

Lbqds:


src/contract/contract.ts
```diff
@@ -265,7 +265,7 @@ export class Contract extends Common {
   }
 
   private static async loadContractStr(sourceFile: SourceFile): Promise<string> {
-    return Common._loadContractStr(sourceFile, [], (code) => Contract.checkCodeType(sourceFile.contractPath, code))
+    return Common._loadContractStr(sourceFile, [sourceFile.contractPath], (code) => Contract.checkCodeType(sourceFile.contractPath, code))
```
Lbqds, line 268: If there is a circular dependency(e.g. contract A depends on contract B, and contract B depends on contract A), it will send duplicated contract code to the full node, we need to add the source contract to the import cache.

Lbqds:
Tests will pass after we make a new full release.

polarker (approved):
LGTM

PR #23, opened 2022-08-23
Support for compiling all contracts and scripts in the project

Lbqds:


src/contract/contract.ts
```diff
@@ -26,165 +26,310 @@ import { node } from '../api'
...
 
-  protected static _artifactsFolder(): string {
-    return './artifacts/'
+export class Project {
+  sourceFiles: SourceFile[]
+  contracts: ArtifactInfo<Contract>[]
+  scripts: ArtifactInfo<Script>[]
+
+  private static instance: Project
+  private static nodeProvider: NodeProvider
```
polarker, line 129: `nodeProvider` can be singleton, but I am not sure about `instance`.
1. we might support subprojects for any source directory, instead of `contracts`
2. one might create multiple projects.

Maybe we could load a project when `fromSource` is called?

src/contract/contract.ts
```diff
@@ -26,165 +26,310 @@ import { node } from '../api'
...
+    return path.startsWith(`./${Project.contractsPath}`)
+      ? path.slice(2)
+      : path.startsWith(Project.contractsPath)
+      ? path
+      : Project.contractsPath + '/' + path
+  }
+
+  contract(path: string): Contract {
+    const contractPath = Project.getContractPath(path)
+    const selected = this.contracts.filter((c) => c.sourceFile.contractPath === contractPath)
```
polarker, line 168: Nit: it seems we could use `find` instead of `filter`

src/contract/contract.ts
```diff
@@ -26,165 +26,310 @@ import { node } from '../api'
...
+          await loadDir(newPath, results)
+        }
+      }
+    }
+    const results: SourceFile[] = []
+    await loadDir(rootPath, results)
+    return results
+  }
+
+  static async load(): Promise<Project> {
```
polarker, line 297: This should be private if we want to call `getInstance()` instead

src/contract/contract.ts
```diff
@@ -26,165 +26,310 @@ import { node } from '../api'
...
+    const sourceFiles = await Project.loadSourceFiles(Project.contractsPath)
+    const contractAndScriptSize = sourceFiles.filter(
+      (f) => f.type === SourceType.Contract || f.type === SourceType.Script
+    ).length
+    if (sourceFiles.length === 0 || contractAndScriptSize === 0) {
+      throw new Error('Project have no source files')
+    }
+    sourceFiles.sort((a, b) => a.type - b.type)
```
polarker, line 305: We assume that the files are sorted before submitted to full node. Is it possible to add a test cover the order somehow?

src/contract/contract.ts
```diff
@@ -26,165 +26,310 @@ import { node } from '../api'
...
+
+  private static async loadSourceFiles(rootPath: string): Promise<SourceFile[]> {
+    const loadDir = async function (dirPath: string, results: SourceFile[]): Promise<void> {
+      const dirents = await fsPromises.readdir(dirPath, { withFileTypes: true })
+      for (const dirent of dirents) {
+        if (dirent.isFile()) {
+          const file = await Project.loadSourceFile(dirPath, dirent.name)
+          results.push(file)
+        } else {
+          const newPath = dirPath + '/' + dirent.name
```
polarker, line 287: Just in case I forget: this will not work on Windows system. We will need another task to support windows

polarker (commented):
Great refactorization!

src/contract/contract.ts
```diff
@@ -26,165 +26,310 @@ import { node } from '../api'
...
+      await Project.load()
     }
-    return fsPromises.writeFile(sourceFile.artifactPath, this.toString())
+    return Project.instance
+  }
+}
+
+export abstract class Artifact {
+  readonly functions: FunctionSig[]
+  readonly warnings: string[]
```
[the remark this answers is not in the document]
Lbqds: Sounds great 👍

src/contract/contract.ts
Lbqds: Another reason I made it a singleton is because the `Contract` depends on the `Project`, and the contract needs to find other contracts by the contract code hash, which is used in `testPublicMethod/testPrivateMethod`.

If we want to support subprojects, I guess we need to add:

1. Add a `Project.fromSource(dirPath)` method used to load the project
2. Add a parameter to the contract `testPublicMethod/testPrivateMethod` methods, so it can get the other contracts by contract code hash.

WDYT?

src/contract/contract.ts
Lbqds: Thanks!

src/contract/contract.ts
Lbqds: I'll try to add a test for this.

src/contract/contract.ts
Lbqds: I used the nodejs path lib at the beginning, but seems it doesn't work with the webpack, and the [path-browserify](https://github.com/browserify/path-browserify) only implements the POSIX functions.

src/contract/contract.ts
Lbqds: I made these fields non-static in the latest commit, and now we can support creating multiple projects.

PR #26, opened 2022-08-29
Support contract type for fields

polarker:


Lbqds (approved):
LGTM 👍, I'll update the bridge contracts.

PR #29, opened 2022-09-03
Monorepo

polarker:


h0ngcha0:
Scanned through the change, the structure looks really nice. The division between web3 and web3-wallet is great too :+1:

PR #30, opened 2022-09-06
Add type id to artifact

Lbqds:
I am currently using regex to get the contract/script type id, maybe we can make the `CompileContract/CompileScript` endpoint return the type id, WDYT? @h0ngcha0 @polarker


polarker:
> I am currently using regex to get the contract/script type id, maybe we can make the `CompileContract/CompileScript` endpoint return the type id, WDYT? @h0ngcha0 @polarker

Agree, how will use the type id? I thought of this, but did not find a use case.

Lbqds:
> > I am currently using regex to get the contract/script type id, maybe we can make the `CompileContract/CompileScript` endpoint return the type id, WDYT? @h0ngcha0 @polarker
> 
> Agree, how will use the type id? I thought of this, but did not find a use case.

It's used to contract deployment. When the user calls `deployer.deploy("Foo")`, the deployer will save the contract deployment result, and the user can get `contractId/txId/blockHash` by `getDeployResult("Foo")`.

polarker:
> It's used to contract deployment. When the user calls `deployer.deploy("Foo")`, the deployer will save the contract deployment result, and the user can get `contractId/txId/blockHash` by `getDeployResult("Foo")`.

I will add this for full node today!

PR #31, opened 2022-09-09
Add command line tool

Lbqds:
I published an npm package for tests:

1. create an empty dir: `mkdir greeter`
2. init the project: `npx alph-cli init`
3. install dependencies: `npm install`
4. start the devnet: `npx alph-cli node start`
5. compile the project: `npx alph-cli compile`
6. run tests: `npx alph-cli test`
7. deploy the contract to the devnet: `npx alph-cli deploy -n devnet`
8. stop the devnet: `npx alph-cli node stop`


Lbqds:
Read for review, now we only support four commands: `init`, `compile`, `test`, and `deploy`.

And I also did some research, truffle does not support typescript very well, for truffle users, you need to use [typechain](https://github.com/dethcrypto/TypeChain) if you want to use typescript, here is an [example](https://github.com/dethcrypto/TypeChain/tree/master/examples/truffle-v5).

Hardhat supports typescript very well. But for ts script execution, we are different from hardhat:

Hardardhat executes ts script by running ts-node in a child process. But we need to compile typescript into javascript first, then execute the js script, which means that the user first needs to execute `npm run build` to compile the typescript file into a javascript file. The reason why we can't use ts-node directly in the child process is that:

For the following code:

```
const deployContract: DeployFunction<Settings> = async (deployer: Deployer, network: Network<Settings>) => {
  const foo = Project.contract('Foo')
  // ....
}
```

`Project.contract` will use `Project.currentProject`, the `currentProject` is undefined in the child process, the user needs to execute `Project.build` to initialize `Project.currentProject` in every typescript file.

So we need to build the project and set the node provider before running the user-defined ts script in the child process, I have tried the following method(inspired by hardhat), first create the `init.js`:

```
async function init() {
  web3.setCurrentNodeProvider('..')
  await Project.build(..)
}

init()
```

then execute `ts-node --require init.js user-defined-ts.ts`, but it doesn't work because the `init()` function is asynchronous, we have to wait for it completed before starting the user-defined ts script, but the top level await does not work for nodejs CommonJS.

So if we want to skip running `npm run build`, maybe we need to support loading the contract/script from the artifact folder then we don't need to dependent the `Project.currentProject`. WDYT?

polarker:
@Lbqds I am thinking of introducing an optional script `init_deploy.ts` with an asynchronous `init` function. If this script file is provided, then we call it. Otherwise, `deployment.ts` call `Project.build()` by default. WDYT?

polarker (approved):
LGTM. I added a few improvements in another PR.

PR #35, opened 2022-09-10
Add name to artifact

Lbqds:
I think we can support `Project.contract(name)` and `Project.script(name)` after this [issue](https://github.com/alephium/alephium/issues/727) solved.


polarker:
> I think we can support `Project.contract(name)` and `Project.script(name)` after this [issue](https://github.com/alephium/alephium/issues/727) solved.

👍  We could also remove the restriction of one contract/script per file

PR #38, opened 2022-09-10
Refactor models

polarker:
* Move the common types from 'contract' to 'api'
* Expose models that are imported from node api


Lbqds (approved):
LGTM 👍

PR #42, opened 2022-09-19
Fix cli

Lbqds:


packages/cli/.npmignore
```diff
@@ -13,7 +13,6 @@ node_modules/
 
 # Config files
 license-header.js
-/tsconfig.json
```
Lbqds, line 16: I removed this as `ts-node` needs tsconfig.json.

packages/cli/scripts/copy-template.js
```diff
@@ -1,8 +1,3 @@
 const fsExtra = require('fs-extra')
-fsExtra.copySync("contracts", "dist/contracts")
-fsExtra.copySync("templates", "dist/templates")
-fsExtra.copySync("devnet-user.conf", "dist/devnet-user.conf")
-fsExtra.copySync(".editorconfig", "dist/.editorconfig")
-fsExtra.copySync(".eslintignore", "dist/.eslintignore")
-fsExtra.copySync(".gitattributes", "dist/.gitattributes")
-fsExtra.copySync("gitignore", "dist/gitignore")
+fsExtra.copySync(".gitignore", "dist/gitignore")
+fsExtra.copySync(".npmignore", "dist/npmignore")
```
Lbqds, line 3: I renamed and copied `.gitignore` and `.npmignore` to dist because `npm publish` will ignore these two files.

PR #62, opened 2022-10-06
Improve signer

polarker:


packages/cli/src/deployment.ts
```diff
@@ -264,7 +264,10 @@ async function createDeployer<Settings = unknown>(
     console.log(`Deploying contract ${taskId}`)
     console.log(`Deployer - group ${signer.group} - ${signer.address}`)
     const result = await contract.transactionForDeployment(signer, params)
-    await signer.submitTransaction(result.unsignedTx)
+    await signer.signAndSubmitUnsignedTx({
+      unsignedTx: result.unsignedTx,
+      signerAddress: (await signer.getSelectedAccount()).address
```
polarker, line 269: A `signerAddress` is required for each transaction because the selected account might have changed since the unsigned tx is built.

Signer will check the signer address before singing, this could provide more security. WDYT?

packages/web3/src/signer/signer.ts
```diff
@@ -318,6 +288,22 @@ export abstract class SignerWithNodeProvider implements SignerProvider {
   abstract signRaw(signerAddress: string, hexString: string): Promise<string>
 }
 
+export abstract class SignerProviderWithMultipleAccounts extends SignerProvider {
```
h0ngcha0, line 291: Is the only non multiple accounts `SingerProvider` the `PrivateKeyWallet`? If that is the case would it make sense to have `SignerProvider` support multiple accounts by default and throw exception when `PrivateKeyWallet` wants to set the selected address? Since the user of the lib would probably always use this one instead. WDYT?

h0ngcha0 (commented):
LGTM :+1: 
Some minor comments

packages/web3/src/signer/signer.ts
polarker: The methods in this class are less popular: `getAccounts, setSelectedAccount`. WC wallet and extension wallet would not support them, so I think it's good to leave them outside `SignerProvider`. WDTY?

packages/web3/src/signer/signer.ts
```diff
@@ -268,31 +245,24 @@ export abstract class SignerWithNodeProvider implements SignerProvider {
...
-      {
-        fromGroup: decoded.fromGroup,
-        toGroup: decoded.toGroup,
-        signerAddress: params.signerAddress,
-        unsignedTx: params.unsignedTx,
-        txId: decoded.unsignedTx.txId
-      },
-      params.submitTx ? params.submitTx : true // we don't consider `alwaysSubmitTx` as the tx might needs multiple signatures
-    )
+    return this.handleSign({
```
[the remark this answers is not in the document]
polarker: Great catch!

packages/web3-wallet/src/node-wallet.ts
```diff
@@ -34,25 +36,12 @@ export class NodeWallet extends SignerWithNodeProvider {
...
-      .getCurrentNodeProvider()
-      .wallets.postWalletsWalletNameChangeActiveAddress(this.walletName, { address: address })
+  async setSelectedAccount(address: string): Promise<void> {
+    await this.nodeProvider.wallets.postWalletsWalletNameChangeActiveAddress(this.walletName, { address: address })
   }
 
-  async getActiveAccount(): Promise<Account> {
-    const activeAddress = await web3.getCurrentNodeProvider().wallets.getWalletsWalletNameAddresses(this.walletName)
+  async getSelectedAccount(): Promise<Account> {
+    const activeAddress = await this.nodeProvider.wallets.getWalletsWalletNameAddresses(this.walletName)
```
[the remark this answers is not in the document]
polarker: Fixed!

packages/web3/src/signer/signer.ts
h0ngcha0: Good point, then I agree with this approach. Thanks!

h0ngcha0 (approved):
LGTM! :100:

PR #64, opened 2022-10-07
Make node provider immutable

polarker:


packages/web3/src/api/index.ts
```diff
@@ -55,22 +55,6 @@ export class NodeProvider {
     this.miners = nodeApi.miners
     this.events = nodeApi.events
   }
-
-  // Have to duplicate the code above due to proxy pattern
-  reset(baseUrl: string, apiKey?: string): void {
-    const nodeApi = initializeNodeApi(baseUrl, apiKey)
-
-    this.wallets = nodeApi.wallets
```
Lbqds, line 63: How about adding the `readonly` to these fields?

Lbqds (approved):
LGTM

packages/web3/src/api/index.ts
polarker: Great one, done!

PR #68, opened 2022-10-11
Refactor signer

polarker:


packages/web3/src/signer/signer.ts
```diff
@@ -141,9 +132,34 @@ assertType<Eq<SignMessageParams, { message: string } & SignerAddress>>()
...
+  signAndSubmitTransferTx(params: SignTransferTxParams): Promise<SignTransferTxResult>
+  signAndSubmitDeployContractTx(params: SignDeployContractTxParams): Promise<SignDeployContractTxResult>
+  signAndSubmitExecuteScriptTx(params: SignExecuteScriptTxParams): Promise<SignExecuteScriptTxResult>
+  signAndSubmitUnsignedTx(params: SignUnsignedTxParams): Promise<SignUnsignedTxResult>
+
+  signUnsignedTx(params: SignUnsignedTxParams): Promise<SignUnsignedTxResult>
+  // The message will be prefixed with 'Alephium Signed Message: ' before signing
+  // so that the resulted signature cannot be reused for building transactions.
+  signMessage(params: SignMessageParams): Promise<SignMessageResult>
+}
```
polarker, line 160: I removed those `sing-only` methods from provider as they are less popular and can be done by `signUnsignedTx`.

The goal is to keep this interface as simple as possible. Please point it out if you think we need to add other methods.

packages/web3/src/signer/signer.ts
```diff
@@ -141,9 +132,34 @@ assertType<Eq<SignMessageParams, { message: string } & SignerAddress>>()
...
+
+export abstract class SignerProviderSimple implements SignerProvider {
```
polarker, line 162: I am not sure what is a better naming for this class

packages/web3/src/signer/signer.ts
```diff
@@ -154,26 +170,30 @@ export abstract class SignerProvider {
...
 
-  async signAndSubmitTransferTx(params: SignTransferTxParams): Promise<SubmissionResult> {
+  async signAndSubmitTransferTx(params: SignTransferTxParams): Promise<SignTransferTxResult> {
     const signResult = await this.signTransferTx(params)
-    return this.submitTransaction(signResult.unsignedTx, signResult.signature)
+    await this.submitTransaction(signResult)
+    return signResult
   }
-  async signAndSubmitDeployContractTx(params: SignDeployContractTxParams): Promise<SubmissionResult> {
+  async signAndSubmitDeployContractTx(params: SignDeployContractTxParams): Promise<SignDeployContractTxResult> {
```
polarker, line 183: More information are returned in all of the submission methods.

packages/web3/src/signer/signer.ts
```diff
@@ -297,31 +298,6 @@ export abstract class SignerProviderWithMultipleAccounts extends SignerProvider
   abstract setSelectedAccount(address: string): Promise<void>
 }
 
-export class SignerProviderWrapper extends SignerProvider {
```
polarker, line 300: This is removed as the signer should always have access (might be indirectly) to a node provider to build transactions.

packages/web3/src/contract/contract.ts
```diff
@@ -822,16 +825,12 @@ export class Contract extends Artifact {
...
     params: Omit<BuildDeployContractTx, 'signerAddress'>
   ): Promise<DeployContractTransaction> {
-    const signerParams = await this.paramsForDeployment({
-      ...params,
-      signerAddress: (await signer.getSelectedAccount()).address
-    })
-    const response = await signer.buildContractCreationTx(signerParams)
-    return fromApiDeployContractUnsignedTx(response)
+    const signerParams = await this.txParamsForDeployment(signer, params)
+    return signer.signAndSubmitDeployContractTx(signerParams)
```
polarker, line 833: This change should make deployment more dev-friendly.

packages/web3/src/contract/contract.ts
```diff
@@ -896,9 +895,12 @@ export class Script extends Artifact {
     return JSON.stringify(object, null, 2)
   }
 
-  async paramsForDeployment(params: BuildExecuteScriptTx): Promise<SignExecuteScriptTxParams> {
+  async txParamsForDeployment(
```
h0ngcha0, line 898: [Nit] 
```suggestion
  async txParamsForExecution(
```

packages/web3/src/signer/signer.ts
h0ngcha0: It should be fine as far as I can see with the new return types :+1:

packages/web3/src/signer/signer.ts
h0ngcha0: Maybe we can call `SignerProvider` as `ISignerProvider` and `SignerProviderSimple` as `SignerProvider` ?

h0ngcha0 (approved):
LGTM!
Only a few minor comments.

packages/web3/src/signer/signer.ts
polarker: `SignerProviderSimple` will not be used for many devs as WC, extension wallet would not extend it. It would be a bit confusing to give it a popular name like `SignerProvider`.

PR #69, opened 2022-10-11
Improve types

polarker:
To work better with JS/TS, this PR uses bigint for amounts. Reasons:
1. operations like `1n + 1` is not allowed in JS
2. the change will make the type of amount more predictable
3. no precision issues

There are some cons too:
1. No scientific notion support: 1e18n is invalid. The neat way I found so far is `10n ** 18n`.
2. A bit tedious to postfix `n` for amount literals. It's less of a problem because the compiler warns if a postfix is missing.

Similar design is made for `web3.js` as far as I have seen.


Lbqds (approved):
LGTM

h0ngcha0 (approved):
LGTM :100:

PR #78, opened 2022-11-18
Fix fetch contract state

Lbqds:
The `Contract.fetchState` depends on `Project.currentProject`, which doesn't work on the browser.

### DIFF
--- a/packages/web3/src/contract/contract.ts
+++ b/packages/web3/src/contract/contract.ts
@@ -618,7 +618,7 @@ export class Contract extends Artifact {
     const state = await web3.getCurrentNodeProvider().contracts.getContractsAddressState(address, {
       group: group
     })
-    return this.fromApiContractState(state)
+    return this._fromApiContractState(state)
   }
 
   override toString(): string {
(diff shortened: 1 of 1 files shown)


packages/web3/src/contract/contract.ts
```diff
@@ -745,7 +745,7 @@ export class Contract extends Artifact {
     }
   }
 
-  private _fromApiContractState(state: node.ContractState): ContractState {
+  fromApiContractState(state: node.ContractState): ContractState {
```
polarker, line 748: `fetchState(address: group)` should work with your fix now, and no need to make this public?

Lbqds: It's useful in some cases, for example, in the bridge UI, when the user selects a token, we don't know whether the token is a local token or a wrapper for a remote token. Now in the bridge UI, we will first get the contract state by the `NodeProvider`, and get the contract type according to the code hash. So if this function is public, we don't need to re-fetch the contract state.

polarker (approved):
LGTM

PR #83, opened 2022-12-26
Fix isHexString

Lbqds:
In the current implementation:

```
isHexString('0x11') is true
isHexString('xxzz') is true
```

### DIFF
--- a/packages/web3/src/utils/utils.ts
+++ b/packages/web3/src/utils/utils.ts
@@ -63,7 +63,7 @@ export function xorByte(intValue: number): number {
 }
 
 export function isHexString(input: string): boolean {
-  return input.length % 2 === 0 && /[0-9a-f]*$/.test(input)
+  return input.length % 2 === 0 && /^[0-9a-fA-F]*$/.test(input)
 }
 
 enum AddressType {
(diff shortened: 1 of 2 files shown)


packages/web3/src/utils/utils.ts
```diff
@@ -63,7 +63,7 @@ export function xorByte(intValue: number): number {
 }
 
 export function isHexString(input: string): boolean {
-  return input.length % 2 === 0 && /[0-9a-f]*$/.test(input)
+  return input.length % 2 === 0 && /^[0-9a-fA-F]*$/.test(input)
```
polarker, line 66: Full node does not support uppercase hexstinrg. If we want to change it, we should change both. Maybe we should support prefix too.

Lbqds: > Full node does not support uppercase hexstinrg.

Do you mean the [Hex](https://github.com/alephium/alephium/blob/55f49f3939cf22b4173b0a1bc0dd92a1e6cf211b/util/src/main/scala/org/alephium/util/Hex.scala#L25)? Just tested it, it supports uppercase hex string.

polarker: Okay, I thought it does not. Then let's merge this first.

PR #86, opened 2023-01-06
Upgrade to full node 1.6.0-rc5

Lbqds:


Lbqds:
It seems that the `npm run audit` failed, but I can't see the error log, and can't download the error log as well.

PR #89, opened 2023-01-16
Refactor signer

polarker:


packages/web3/src/signer/signer.ts
```diff
@@ -161,11 +67,23 @@ export interface SignerProvider {
   signMessage(params: SignMessageParams): Promise<SignMessageResult>
 }
 
+// Abstraction for interactive signer (e.g. WalletConnect instance, Extension wallet object)
+export interface InteractiveSignerProvider<ConnectOptions extends { chainGroup?: number }> extends SignerProvider {
+  // @param chainGroup - specify whether to use addresses from a specific group
+  connect(opt?: ConnectOptions): Promise<void>
```
polarker, line 73: Should it be renamed as `enable` as `Argent-X`?

packages/web3-wallet/src/hd-wallet.ts
```diff
@@ -60,3 +67,67 @@ export function getHDWalletPath(addressIndex: number): string {
...
+    let usedAddressIndex = -1
+    for (const account of this._accounts.values()) {
+      if ((targetGroup === undefined || account.group == targetGroup) && account.addressIndex > usedAddressIndex) {
+        usedAddressIndex = account.addressIndex
+      }
+    }
+    return usedAddressIndex + 1
+  }
+
+  deriveNewAccount(targetGroup?: number): HDWalletAccount {
```
h0ngcha0, line 97: Q: When we derive a new account, we might have already generated/persisted a few accounts from before. Would it be good to allow passing in the `addressIndex`es that are already used?
IIUC, the cached accounts here is not persisted, right?

packages/web3/src/signer/signer.ts
h0ngcha0: Sure :+1:

packages/web3-wallet/src/hd-wallet.test.ts
```diff
@@ -50,29 +55,75 @@ describe('HD wallet', () => {
...
       '51df72c50e1b392805c3f530d2dece7e0c1842f827a579f7cd24907a6c4ff533'
     )
-    expect(deriveHDWalletPrivateKeyForGroup(testMnemonic, 3, 100, 'Alephium')).toEqual(
+    expect(deriveHDWalletPrivateKeyForGroup(testMnemonic, 3, 100, 'Alephium')[0]).toEqual(
       '04528b7736eab20cbde90f0d2f0cb3a99481dfe92664757646077ae7851e4314'
     )
   })
+
+  it('should derive account', async () => {
+    const wallet = new HDWallet(testMnemonic, undefined, undefined, testMnemonic)
```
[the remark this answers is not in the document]
polarker: Great catch!

packages/web3-wallet/src/hd-wallet.ts
polarker: Yes, it's done in this line:
```
    const fromAddressIndex = this.getNextFromAddressIndex(targetGroup)
```
The derivation will start from the smallest available index.

packages/web3/src/signer/signer.ts
polarker: What should we rename 'ConnectOptions'? `EnableOptions` sounds good?

polarker: Done!

packages/web3-wallet/src/hd-wallet.ts
polarker: Yes, that's totally right. This is an in-memory wallet designed for non-critical use cases, for example, contract deployment. For productions wallet (e.g. extension wallet), those derivation functions should be used instead. This wallet is not generic or flexible enough for production wallets. Same for `PrivateKeyWallet`. I will add a comment on this.

WDYT?

polarker: Btw, the derived addresses are added to `_accounts` if it's derived with `deriveAndAddNewAccount`.

h0ngcha0: Got it, sounds great :+1:

polarker: Just simplified HDWallet to remove `deriveNewAccount` and keep `deriveAndAddNewAccount` only.

PR #94, opened 2023-01-31
Token standard

Lbqds:
* Add `Token` interface, all token contracts should implement the token interface
* Support importing from external libs, it works similarly to [truffle](https://trufflesuite.com/docs/truffle/how-to/package-management-via-npm/)

Now users can import the `Token` interface:

```
import "std/token_interface.ral"

Contract Token() implements IToken {
  // ...
}
```

You can refer to the [dApp template](https://github.com/alephium/alephium-web3/blob/token-standard/packages/cli/templates/base/contracts/token.ral) for more info.

But there is one place that is not very user-friendly. For example,  if there is an external lib that has two contract files:

1. contract `external-lib/contracts/a.ral`:
    ```
    Interface A {
      pub fn a() -> () 
    }
    ```
2. contract `external-lib/contracts/b.ral`:
    ```
    Interface B extends A {
      pub fn b() -> ()
    }
    ```

We also need to import the `a.ral` if the contract only depends on interface `B`:

```
import "external-lib/contracts/a.ral"  // The compiler emits interface `A` not found error if we don't import this file.
import "external-lib/contracts/b.ral"

Contract Foo() implements B {
  pub fn a() -> () {}
  pub fn b() -> () {}
}
```

We compile a project as a whole, so it is not necessary to import files from the same project. I'm not sure if there is a good way to solve this.


contracts/token_test.ral
```diff
@@ -0,0 +1,24 @@
+import "std/token_interface.ral"
```
polarker, line 1: Let's remove the postfix in the example contracts.

packages/cli/templates/base/contracts/token.ral
```diff
@@ -1,9 +1,17 @@
+import "std/token_interface.ral"
```
polarker, line 1: same here

polarker (approved):
Great work 👍

PR #96, opened 2023-02-10
Introduce code generator

Lbqds:


.generated/Add.ts
```diff
@@ -0,0 +1,436 @@
...
+  a: bigint;
+  b: bigint;
+}
+
+export type _AddFields = {
+  sub: string;
+  result: bigint;
+};
+
+export type _AddState = _AddFields & Omit<ContractState, "fields">;
```
polarker, line 50: 1. Let's use namespace
2. All events should use `Event` as postfix
```
export declare namespace Add {
  export type Fields = {
    sub: string;
    result: bigint;
  }

  export type State = Fields & Omit<ContractState, "fields">

  export type AddEvent = {
    blockHash: string;
    txId: string;
    eventIndex: number;
    x: bigint;
    y: bigint;
  }

  export type Add1Event = {
    blockHash: string;
    txId: string;
    eventIndex: number;
    a: bigint;
    b: bigint;
  }
}
```

.generated/Add.ts
```diff
@@ -0,0 +1,436 @@
...
+    sub: string,
+    result: bigint,
+    _extraParams?: {
+      initialAttoAlphAmount?: bigint;
+      initialTokenAmounts?: Token[];
+      issueTokenAmount?: bigint;
+      gasAmount?: number;
+      gasPrice?: bigint;
+    }
+  ): Promise<Add> {
```
polarker, line 166: How about:
```
  static async deploy(
    signer: SignerProvider,
    fields: Add.Fields,
    deployParams?: {
      initialAttoAlphAmount?: bigint;
      initialTokenAmounts?: Token[];
      issueTokenAmount?: bigint;
      gasAmount?: number;
      gasPrice?: bigint;
    }
  ): Promise<Add> {
  ```

File: .generated/Add.ts [hunk 3]
```diff
@@ -0,0 +1,436 @@
...
+      _deployResult.contractId,
+      _deployResult.fromGroup,
+      _deployResult
+    );
+  }
+
+  static connect(
+    address: string,
+    deployResult?: SignDeployContractTxResult
+  ): Add {
```
polarker, line 186: 1. deployResult can be removed from the class
2. TypedChain uses `attach`, which sounds better
```
static attach(
    address: string,
  ): Add {
```

.generated/Add.ts
```diff
@@ -0,0 +1,436 @@
...
+  async fetchState(): Promise<_AddState> {
+    const state = await Add.contract.fetchState(this.address, this.groupIndex);
+    return {
+      ...state,
+      sub: state.fields["sub"] as string,
+      result: state.fields["result"] as bigint,
+    };
+  }
+
+  private decodeAdd(event: node.ContractEvent): AddAdd {
```
polarker, line 201: add postfix `decodeAddEvent`

.generated/Add.ts
```diff
@@ -0,0 +1,436 @@
...
+    return {
+      blockHash: event.blockHash,
+      txId: event.txId,
+      eventIndex: event.eventIndex,
+      x: fields["x"] as bigint,
+      y: fields["y"] as bigint,
+    };
+  }
+
+  subscribeAdd(
```
polarker, line 217: add postfix `subscribeAddEvent`

.generated/Add.ts
```diff
@@ -0,0 +1,436 @@
...
+      "addPrivate",
+      _testParams
+    );
+    return {
+      ..._testResult,
+      returns: _testResult.returns as [[bigint, bigint]],
+    };
+  }
+
+  async addCall(
```
polarker, line 410: maybe `callAddMethod`

.generated/Add.ts
```diff
@@ -0,0 +1,436 @@
...
+      tokens: asset?.tokens,
+    };
+    return Add.contract.toState(
+      { sub: sub, result: result },
+      newAsset,
+      address
+    );
+  }
+
+  static async testAdd(
```
polarker, line 343: maybe `testAddMethod`

.generated/Add.ts
```diff
@@ -0,0 +1,415 @@
...
+    initFields: Fields,
+    deployParams?: {
+      initialAttoAlphAmount?: bigint;
+      initialTokenAmounts?: Token[];
+      issueTokenAmount?: bigint;
+      gasAmount?: number;
+      gasPrice?: bigint;
+    }
+  ): Promise<SignDeployContractTxResult & { instance: AddInstance }> {
+    const deployResult = await artifact.deploy(signer, {
```
polarker, line 66: We should remove `artifact.deploy` and convert directly from node result, so that there will be only one way to deploy contracts.

We could refactor the conversion code as helper functions in `contract.ts` so we could use them in the generated code.

.generated/Add.ts
```diff
@@ -0,0 +1,415 @@
...
+      tokens: testParams?.initialAsset?.tokens,
+    };
+    const _testParams = {
+      ...testParams,
+      testMethodIndex: 0,
+      testArgs: args,
+      initialFields: initFields,
+      initialAsset: initialAsset,
+    };
+    const testResult = await artifact.testPublicMethod("add", _testParams);
```
polarker, line 124: 1. We need to convert the result so that `returns` and `events` have proper types
2. We could also convert the result directly from `node....`, like `deploy` function.

packages/cli/cli_internal.ts
```diff
@@ -71,9 +71,11 @@ program
   .description('compile the project')
   .option('-c, --config <config-file>', 'project config file (default: alephium.config.{ts|js})')
   .option('-n, --network <network-type>', 'network type')
+  .option('-g, --generate', 'generate typescript code by contract artifacts')
+  .option('-o, --outDir <out-dir>', 'the output directory')
```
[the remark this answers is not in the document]
Lbqds: Sounds good, will update.

.generated/Add.ts
Lbqds: Sorry, I don't understand what you mean by `convert directly from node result`?

.generated/Add.ts
Lbqds: For returns, it has been converted to a proper typescript type. For ralph array types such as `[[U256; 3]; 2]`, it will be converted to typescript `[[bigint, bigint, bigint], [bigint, bigint, bigint]]` instead of `bigint[][]`. But since only one value is returned here, the return type should be `[bigint, bigint]` rather than `[[bigint, bigint]]`, will fix this, thanks!

Lbqds: I found a problem for convert event type, for example:

```
Contract Foo(bar: Bar) {
  fn foo() -> () {
    bar.bar()
  }
}

Contract Bar() {
  event E(v: U256)
  pub fn bar() -> () {
    emit E(0)
  }
}
```

The event is emitted from contract `Bar` when calling `Foo.foo`, in this case, we can not convert the event type.

packages/cli/src/types.ts
```diff
@@ -107,32 +111,42 @@ export async function getEnv<Settings = unknown>(configFileName?: string): Promi
...
-export type RunScriptParams = Omit<BuildExecuteScriptTx, 'signerAddress'>
-
 export interface Deployer {
   provider: NodeProvider
   account: Account
 
-  deployContract(contract: Contract, params: DeployContractParams, taskTag?: string): Promise<DeployContractResult>
-  runScript(script: Script, params: RunScriptParams, taskTag?: string): Promise<RunScriptResult>
+  deployContract<T, P extends Fields>(
+    constractFactory: ContractFactory<T, P>,
```
Lbqds, line 137: I introduced the `ContractFactory` for deploying contracts more conveniently.

packages/cli/src/types.ts
```diff
@@ -107,32 +111,42 @@ export async function getEnv<Settings = unknown>(configFileName?: string): Promi
...
+    params: DeployContractParams<P>,
+    taskTag?: string
+  ): Promise<DeployContractResult<T>>
+
+  runScript<P extends Fields>(
+    executeFunc: (singer: SignerProvider, params: ExecuteScriptParams<P>) => Promise<ExecuteScriptResult>,
```
Lbqds, line 143: Not sure if we want to introduce `ExecutableScript` like `ContractFactory`.

PR #99, opened 2023-02-15
Deploy with private key

Lbqds:


packages/cli/src/deployment.ts
```diff
@@ -183,7 +183,8 @@ async function needToRetry(
   }
   const currentTokens = tokens ? tokens : {}
   const previousTokens = previous.tokens ? previous.tokens : {}
-  const sameWithPrevious = attoAlphAmount === previous.attoAlphAmount && recordEqual(currentTokens, previousTokens)
+  const sameWithPrevious =
+    attoAlphAmount?.toString() === previous.attoAlphAmount?.toString() && recordEqual(currentTokens, previousTokens)
```
Lbqds, line 187: I found a strange bug, even if `attoAlphAmount` and `previous.attoAlphAmount` are equal, it will return false here, the reason is that `typeof attoAlphAmount` is bigint but `typeof previous.attoAlphAmount` is string, I don’t know why.

packages/cli/src/deployment.ts
```diff
@@ -200,7 +201,7 @@ async function needToDeployContract(
     return true
   }
   // previous !== undefined if retry is false
-  return previous!.issueTokenAmount !== issueTokenAmount
+  return previous!.issueTokenAmount?.toString() !== issueTokenAmount?.toString()
```
Lbqds, line 204: Same here.

packages/cli/src/deployment.ts
polarker: I will check

polarker (commented):
LGTM, two minor comments

PR #102, opened 2023-02-16
Fix decode contract created system event

Lbqds:
In full node 1.7.0, the `ContractCreatedEvent` has an optional field `parentAddress`, this PR fix decodes `ContractCreatedEvent`.


polarker (approved):
Thanks 👍

PR #107, opened 2023-02-21
Fix subscribe system events

Lbqds:
We use special contract addresses for system events, we don't need to generate subscribe system events code for contract instances.


polarker:
Thanks, great catch 👍

PR #111, opened 2023-02-23
Move code to format ALPH

h0ngcha0:


nop33:
@h0ngcha0 you might be interested to get my latest code (still not merged) that takes care of displaying tokens as well, not just ALPH: https://github.com/alephium/js-sdk/pull/145

PR #112, opened 2023-02-23
Improve wallet initialization

polarker:


packages/web3-wallet/src/hd-wallet.test.ts
```diff
@@ -110,7 +110,7 @@ describe('HD wallet', () => {
   })
 
   it('should derive account', async () => {
-    const wallet = new HDWallet(testMnemonic, undefined, undefined, undefined, 'Alephium')
+    const wallet = new HDWallet({ mnemonic: testMnemonic, passphrase: 'Alephium' })
```
polarker, line 113: This seems more dev friendly than the previous version

PR #123, opened 2023-02-28
Introduce simple API retry

polarker:


packages/web3/src/api/utils.ts
```diff
@@ -16,10 +16,24 @@ You should have received a copy of the GNU Lesser General Public License
...
   } else {
     return response.data
   }
 }
+
+export async function retryFetch(...fetchParams: Parameters<typeof fetch>): ReturnType<typeof fetch> {
+  const retry = async (retryNum: number): ReturnType<typeof fetch> => {
+    const response = await fetch(...fetchParams)
+    if (response.status === 429 && retryNum < 5) {
+      return await retry(retryNum + 1)
```
Lbqds, line 33: Do we need to wait for some time before starting the next request?

Lbqds (approved):
LGTM, only one question.

packages/web3/src/api/utils.ts
polarker: Great catch. I had that in mind but forgot it somehow 😓

Lbqds (approved):

h0ngcha0 (approved):
LGTM :+1:

PR #141, opened 2023-03-28
Update NFT and NFT Collection Interfaces

h0ngcha0:


packages/web3/std/nft_collection_interface.ral
```diff
@@ -1,8 +1,5 @@
 Interface INFTCollection {
-   event Minted(minter: Address, tokenIndex: U256, tokenId: ByteVec)
-
-   pub fn getName() -> ByteVec
-   pub fn getSymbol() -> ByteVec
+   pub fn getCollectionUri() -> ByteVec
```
polarker, line 2: Let's add a standard for the collection metadata in the comment?

packages/web3/std/nft_collection_interface.ral
```diff
@@ -1,8 +1,5 @@
 Interface INFTCollection {
-   event Minted(minter: Address, tokenIndex: U256, tokenId: ByteVec)
-
-   pub fn getName() -> ByteVec
-   pub fn getSymbol() -> ByteVec
```
polarker, line 5: Maybe we should keep these two most important fields on-chain so that they are totally immutable?

polarker: We decided not to having these two following EIP1155.

packages/web3/std/nft_collection_interface.ral
h0ngcha0: Done :+1:

Issue #150, opened 2023-04-03
Fix circular dependencies


nop33:
I noticed some circular deps in the web3-react package when running the build script:

```
(!) Circular dependencies
src/components/AlephiumConnect.tsx -> src/components/ConnectModal/index.tsx -> src/components/AlephiumConnect.tsx
src/components/AlephiumConnect.tsx -> src/components/ConnectModal/index.tsx -> src/components/Common/Modal/index.tsx -> src/components/AlephiumConnect.tsx
src/components/AlephiumConnect.tsx -> src/components/ConnectModal/index.tsx -> src/components/Pages/Connectors/index.tsx -> src/components/AlephiumConnect.tsx
...and 7 more
```

nop33:
I have already fixed it! I'll make a PR in a sec!

PR #151, opened 2023-04-03
Fix circular imports

nop33:
While trying to use a local version of the `web3-react` package in the [nft-prototype project](https://github.com/alephium/alephium-nft) so that I can work on the WalletConnect compatibility with the desktop wallet, I've been encountering an error where `React` is `null`. I thought this might be related to the circular imports #150  that I noticed in the package. It wasn't related to that, I am still facing this error. Nevertheless, this PR fixes the circular imports and improves some typing.

### DIFF
--- a/packages/web3-react/src/components/AlephiumConnect.tsx
+++ b/packages/web3-react/src/components/AlephiumConnect.tsx
@@ -15,7 +15,7 @@ GNU Lesser General Public License for more details.
 You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
-import React, { createContext, useEffect, useState } from 'react'
+import React, { useEffect, useState } from 'react'
 
 import defaultTheme from '../styles/defaultTheme'
 
@@ -23,47 +23,12 @@ import AlephiumConnectModal from '../components/ConnectModal'
 import { ThemeProvider } from 'styled-components'
 import { Account, KeyType, SignerProvider } from '@alephium/web3'
 import { Theme, Mode, CustomTheme } from '../types'
-
-export const routes = {
-  CONNECTORS: 'CONNECTORS',
-  PROFILE: 'PROFILE',
-  CONNECT: 'CONNECT'
-}
-
-type Connector = any
-type Error = string | React.ReactNode | null
-
-type ContextValue = {
-  open: boolean
-  setOpen: React.Dispatch<React.SetStateAction<boolean>>
-  route: string
-  setRoute: React.Dispatch<React.SetStateAction<string>>
-  errorMessage: Error
-  connector: Connector
-  setConnector: React.Dispatch<React.SetStateAction<Connector>>
-  account?: Account
-  setAccount: React.Dispatch<React.SetStateAction<Account | undefined>>
-  displayAccount?: (account: Account) => string
-  signerProvider?: SignerProvider
-  setSignerProvider: React.Dispatch<React.SetStateAction<SignerProvider | undefined>>
-  addressGroup?: number
-  keyType?: KeyType
-  network?: string
-  theme: Theme
-  setTheme: React.Dispatch<React.SetStateAction<Theme>>
-  mode: Mode
-  setMode: React.Dispatch<React.SetStateAction<Mode>>
-  customTheme: CustomTheme
-  setCustomTheme: React.Dispatch<React.SetStateAction<CustomTheme>>
-}
-
-const Context = createContext<ContextValue | null>(null)
-
-export const useContext = () => {
-  const context = React.useContext(Context)
-  if (!context) throw Error('AlephiumConnect Hook must be inside a Provider.')
-  return context
-}
+import { routes } from './Common/Modal'
+import {
+  AlephiumConnectContext,
+  AlephiumConnectContextValue,
+  useAlephiumConnectContext
+} from '../contexts/alephiumConnect'
 
 type AlephiumConnectProviderProps = {
   useTheme?: Theme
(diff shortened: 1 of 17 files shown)


packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -15,55 +15,16 @@ GNU Lesser General Public License for more details.
...
-  setCustomTheme: React.Dispatch<React.SetStateAction<CustomTheme>>
-}
-
-const Context = createContext<ContextValue | null>(null)
-
-export const useContext = () => {
-  const context = React.useContext(Context)
-  if (!context) throw Error('AlephiumConnect Hook must be inside a Provider.')
-  return context
-}
```
nop33, line 66: Moving all these definitions to a new file fixes the circular import 💫

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -95,10 +57,10 @@ export const AlephiumConnectProvider: React.FC<AlephiumConnectProviderProps> = (
   const [customTheme, setCustomTheme] = useState<CustomTheme>(useCustomTheme ?? {})
 
   const [open, setOpen] = useState<boolean>(false)
-  const [connector, setConnector] = useState<string>('')
+  const [connectorId, setConnectorId] = useState<AlephiumConnectContextValue['connectorId']>('')
   const [route, setRoute] = useState<string>(routes.CONNECTORS)
   const [account, setAccount] = useState<Account>()
-  const [errorMessage, setErrorMessage] = useState<Error>('')
+  const [errorMessage, setErrorMessage] = useState<AlephiumConnectContextValue['errorMessage']>('')
```
nop33, line 63: This way we don't need to redefine these types!

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -110,8 +72,8 @@ export const AlephiumConnectProvider: React.FC<AlephiumConnectProviderProps> = (
     setOpen,
     route,
     setRoute,
-    connector,
-    setConnector,
+    connectorId,
+    setConnectorId,
```
nop33, line 76: It was a bit confusing calling this variable `connector` while its value is just the `id` property of the `Connector` type. I think it's clearer this way!

packages/web3-react/src/components/ConnectModal/ConnectWithWalletConnect.tsx
```diff
@@ -17,15 +17,15 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 import { Container } from './ConnectWithInjector/styles'
 import { useConnect } from '../../hooks/useConnect'
 
 let _init = false
 
 const ConnectWithWalletConnect: React.FC = () => {
-  const context = useContext()
-  const [error, setError] = useState<string | undefined>(undefined)
+  const context = useAlephiumConnectContext()
+  const [error, setError] = useState<string>()
```
nop33, line 28: In React, both of these result to initializing with `undefined`!

packages/web3-react/src/components/ConnectModal/index.tsx
```diff
@@ -43,11 +43,11 @@ const ConnectModal: React.FC<{
...
-  const pages: any = {
-    connectors: <Connectors />,
-    connect: <ConnectUsing connectorId={context.connector} />,
-    profile: <Profile />
-  }
+  const pages: Page[] = [
+    { id: 'CONNECTORS', content: <Connectors /> },
+    { id: 'CONNECT', content: <ConnectUsing connectorId={context.connectorId} /> },
+    { id: 'PROFILE', content: <Profile /> }
+  ]
```
nop33, line 50: This way we have better typing, I hope you like this change!

packages/web3-react/src/components/Pages/Connectors/index.tsx
```diff
@@ -34,29 +34,14 @@ import {
...
-      if (a) return x
-      return null
-    })
-    if (hasWalletLogo.length === 0) return null
-
-    const foundInjector = wallets.filter((wallet: any) => wallet.installed && wallet.name === hasWalletLogo[0])[0]
-
-    return foundInjector
-  }
-
```
nop33, line 59: I moved this function outside of the component so that it's easier to read it and so that it doesn't get redefined every time this component renders :)

packages/web3-react/src/constants/supportedConnectors.tsx
```diff
@@ -100,8 +78,7 @@ if (typeof window != 'undefined') {
         qrCode: <Logos.WalletConnect background={true} />
       },
       logoBackground: 'var(--ck-brand-walletConnect)',
-      scannable: true,
-      defaultConnect: () => {}
```
nop33, line 104: Since this prop is optional, and since there's nothing to initialize with, I removed it to solve the error

packages/web3-react/src/constants/supportedConnectors.tsx
```diff
@@ -15,34 +15,12 @@ GNU Lesser General Public License for more details.
...
-  appUrls?: { [key: string]: string }
-  extensionIsInstalled?: () => any
-  defaultConnect?: () => any
-}[] = []
+let supportedConnectors: Connector[] = []
 
 if (typeof window != 'undefined') {
-  interface IDictionary {
-    [index: string]: string
-  }
```
nop33, line 44: This was not used, so I removed it

packages/web3-react/src/constants/supportedConnectors.tsx
```diff
@@ -15,34 +15,12 @@ GNU Lesser General Public License for more details.
...
-    appIcon?: ReactNode
-    mobile?: ReactNode
-  }
-  logoBackground?: string
-  scannable?: boolean
-  extensions?: { [key: string]: string }
-  appUrls?: { [key: string]: string }
-  extensionIsInstalled?: () => any
-  defaultConnect?: () => any
-}[] = []
```
nop33, line 39: I extracted this into a type in the types file so that it can be reused across the project!

Lbqds (approved):
LGTM, lots of great improvements, thanks!

PR #153, opened 2023-04-04
Add wallet connect support

Lbqds:


nop33 (approved):
LGTM 👍  There are a few changes I would like to suggest, like using different array functions to find an element in an array and stuff like that, but I don't want to delay the process, code improvements can come later, let's make it work first!

packages/walletconnect/test/shared/WalletClient.ts
```diff
@@ -93,7 +94,7 @@ export class WalletClient {
 
   constructor(provider: WalletConnectProvider, opts: Partial<WalletClientOpts>) {
     this.provider = provider
-    this.networkId = opts?.networkId ?? -1
+    this.networkId = opts?.networkId ?? ''
```
polarker, line 97: Maybe we should restrict `networkId` to the three values (mainnet, testnet, devnet) only?

packages/web3-react/src/components/ConnectModal/ConnectWithWalletConnect.tsx
```diff
@@ -0,0 +1,51 @@
...
+    networkId: context.network ?? ''
+  })
+
+  useEffect(() => {
+    if (_init) {
+      return
+    }
+
+    _init = true
+    connect().catch((err) => setError(`${err}`))
```
polarker, line 41: Let's set `_init = true` only when connect succeeds with `connect().then(...)`?

packages/web3-react/src/contexts/alephiumConnect.tsx
```diff
@@ -0,0 +1,56 @@
...
+  setConnectorId: React.Dispatch<React.SetStateAction<Connector['id']>>
+  account?: Account
+  setAccount: React.Dispatch<React.SetStateAction<Account | undefined>>
+  displayAccount?: (account: Account) => string
+  signerProvider?: SignerProvider
+  setSignerProvider: React.Dispatch<React.SetStateAction<SignerProvider | undefined>>
+  addressGroup?: number
+  keyType?: KeyType
+  network?: string
+  networkId?: number
```
polarker, line 41: This can be removed now?

packages/web3-react/src/hooks/useAccount.tsx
```diff
@@ -17,14 +17,18 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+import { useAlephiumConnectContext } from '../contexts/alephiumConnect'
 import { KeyType } from '@alephium/web3'
 
 export function useAccount(onDisconnected?: () => Promise<void>) {
-  const context = useContext()
+  const context = useAlephiumConnectContext()
 
   useEffect(() => {
     const handler = async () => {
+      if (context.connectorId === 'walletConnect') {
```
polarker, line 28: Just to confirm: this means we don't support auto-connect for WalletConnect right now?

packages/web3-react/src/types.ts
```diff
@@ -15,6 +15,28 @@ GNU Lesser General Public License for more details.
...
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
+import { ReactNode } from 'react'
+
 export type Theme = 'auto' | 'web95' | 'retro' | 'soft' | 'midnight' | 'minimal' | 'rounded' | 'nouns'
 export type Mode = 'light' | 'dark' | 'auto'
 export type CustomTheme = any // TODO: define type
+
+export type Connector = {
+  id: string
```
polarker, line 25: Should we introduce a type for this id too?

polarker (commented):
LGTM, just a few nits comments. Ready to merge once they are resoved.

packages/web3-react/src/components/Pages/Connectors/index.tsx
```diff
@@ -107,14 +92,10 @@ const Wallets: React.FC = () => {
               if (!info) return null
 
               let logos = info.logos
+              let name = info.id === 'walletConnect' ? 'WalletConnect' : info.name ?? connector.name
```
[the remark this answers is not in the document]
nop33: > The previous name is Wallet Connect.

I couldn't find many references of "Wallet Connect" online. 

I can change the code back to its more readable form, no problem.

packages/web3-react/src/types.ts
nop33: You mean an enum or sth similar? Sounds like a good idea 👍

packages/web3-react/src/components/Pages/Connectors/index.tsx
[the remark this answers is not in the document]
nop33: updated!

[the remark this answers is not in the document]
polarker: > I couldn't find many references of "Wallet Connect" online.

If it's on purpose change, then I am fine with it :)

packages/web3-react/src/hooks/useAccount.tsx
Lbqds: Yes, because I'm not sure how we should support automatic connection via WalletConnect. I'll look into how to support this later.

packages/web3-react/src/types.ts
polarker: Yes, but I am not sure whether it's doable due to injected connector

packages/web3-react/src/components/ConnectModal/ConnectWithInjector/index.tsx
```diff
@@ -92,7 +92,7 @@ const ConnectWithInjector: React.FC<{
   const { connect } = useConnect({
     chainGroup: context.addressGroup,
     keyType: context.keyType,
-    networkId: context.network
+    networkId: context.network ?? ''
```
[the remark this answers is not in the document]
nop33: Ah, there is already one in the CLI! Nice!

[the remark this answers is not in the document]
nop33: I would suggest that we export an array with the valid values so that we can do validation when needed, and also generate a type out of the array (so that we only define the valid list of values once). Here's how we can do that:

```ts
export const networkIds = ['mainnet', 'testnet', 'devnet'] as const
export type NetworkId = (typeof networkIds)[number]
```

We can update the `NetworkType` type from the CLI tool to this and use it everywhere else. WDYT?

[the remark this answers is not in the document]
Lbqds: I agree with what you said, but I think we should add this to @alephium/web3. I'm working on that now.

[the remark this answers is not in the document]
polarker: That's a great idea 👍 We use similar trick in WalletConnectProvider

packages/web3-react/src/hooks/useAccount.tsx
polarker: Np, I will assign the task to @ross-weir later since he has been looking into this for quit some time and got blocked by us for a while.

PR #155, opened 2023-04-04
Better network ID typing

nop33:


polarker:
Thanks for the PR. I asked Muchen to make such changes as I thought Muchen was taking care of that PR.

As discussed with him, we will likely to introduce this type to `web3` package as it's being used in almost every other packages now.

nop33:
ok, thanks for the feedback!

PR #156, opened 2023-04-04
Resolve wallet connect comments

Lbqds:


packages/web3/src/utils/utils.ts
```diff
@@ -27,6 +27,9 @@ import djb2 from './djb2'
 import { KeyType } from '../signer'
 import { HexString } from '../contract'
 
+export const networkIds = ['mainnet', 'testnet', 'devnet'] as const
+export type NetworkId = (typeof networkIds)[number]
```
Lbqds, line 31: The code comes from @nop33, thanks!

polarker (approved):
LGTM

packages/web3-react/src/components/ConnectModal/ConnectWithWalletConnect.tsx
```diff
@@ -29,16 +29,19 @@ const ConnectWithWalletConnect: React.FC = () => {
...
   useEffect(() => {
     if (_init) {
       return
     }
 
-    _init = true
-    connect().catch((err) => setError(`${err}`))
+    connect()
+      .then(() => {
+        _init = true
```
nop33, line 42: What happens if `error` already has a value before `_init` is set to true? I think we should add the following line inside this function, otherwise, in this particular corner-case, the text below will never show the correct value

```ts
setError(undefined)
```

polarker: Let's leave this comment to this issue: https://github.com/alephium/alephium-web3/issues/154
I feel that this implementation would have to be rewritten soon, so let's spend too much time for now.

Lbqds: Great catch, thanks!

polarker: I thought it's gonna be a big change due to the global variable `_init` and local react state (which might be tricky)

polarker:
Merging now. I have an appointment in the afternoon, so I'd like to release soon.

PR #160, opened 2023-04-05
Load deployments by network id

Lbqds:
Generate the loadDeployments function to facilitate users in obtaining contract deployment results. 

Here is the code generated for the DEX: https://github.com/alephium/alephium-dex/blob/test-load-deployments/artifacts/ts/deployments.ts


packages/cli/src/codegen.ts
```diff
@@ -405,8 +405,109 @@ function genContracts(outDir: string, artifactDir: string, exports: string[]) {
...
+  return path.relative(fromPath, deploymentFilePath)
+}
+
+export async function genLoadDeployments(config: Configuration) {
+  const tsPath = path.join(config.artifactDir ?? DEFAULT_CONFIGURATION_VALUES.artifactDir, 'ts')
+  const deploymentsPath = networkIds.map((n) => [n, getRelativePath(config, n, tsPath)])
+  const deploymentsExists = deploymentsPath.filter(([, filePath]) => filePath !== undefined)
+  const imports = deploymentsExists
+    .map(([networkId, relativePath]) => {
+      return `import { default as ${networkId}Deployments } from '${relativePath}'`
```
Lbqds, line 475: We need to import the JSON files because the front-end cannot dynamically read files based on paths.

packages/cli/src/codegen.ts
```diff
@@ -504,16 +504,24 @@ export async function genLoadDeployments(config: Configuration) {
 
     ${genDeploymentsType(allDeployments)}
 
-    export function loadDeployments(networkId: NetworkId, deployerAddress?: string): Deployments | undefined {
+    export function loadDeployments(networkId: NetworkId, deployerAddress?: string): Deployments {
```
Lbqds, line 507: To keep consistency, I made all of them throw exceptions

polarker (approved):
Great work 💯

PR #165, opened 2023-04-12
Add the `FungibleTokenUnimplemented` contract

Lbqds:


packages/cli/cli_internal.ts
```diff
@@ -94,7 +94,13 @@ program
       }
       web3.setCurrentNodeProvider(nodeUrl)
       const cwd = path.resolve(process.cwd())
-      await Project.build(config.compilerOptions, cwd, config.sourceDir, config.artifactDir)
+      await Project.build(
+        config.compilerOptions,
+        cwd,
+        config.sourceDir,
+        config.artifactDir,
+        DEFAULT_CONFIGURATION_VALUES.nodeVersion
```
polarker, line 102: It seems possible that the configured version is different from the actually full node version?

We perhaps should always get the version via full node API, instead of configuring it.

.github/workflows/openapi.yml
```diff
@@ -22,4 +22,3 @@ jobs:
       # Wait for this issue to be resolved: https://github.com/acacode/swagger-typescript-api/issues/370
       - run: npm run update-schemas && test -z "$(git status --porcelain)"
       - run: npm run build
-      - run: npm run codegen && test -z "$(git status --porcelain)"
```
polarker, line 25: `openapi.yml` might run after `test.yml`, so we may still need to keep this line

polarker (approved):
LGTM!

.github/workflows/openapi.yml
polarker: Sorry, I did not notice that the line is checking codegen. All good 👍

PR #167, opened 2023-04-21
Add getCollectionId to INFT interface

h0ngcha0:


polarker:
> > @h0ngcha0 Why do we suddenly need to update the lock file?
> 
> The chain of events are:
> 
> 1. I used v8 version of pnpm locally, which updated  the lock file. The new lock file doesn't work for CI ([1](https://github.com/alephium/alephium-web3/actions/runs/4760180887/jobs/8461537469), [2](https://github.com/alephium/alephium-web3/actions/runs/4760180885/jobs/8460241596)) since some of the actions only support pnpm v7
> 2. I downgraded the pnpm version and re-generated the pnpm lock file.
> 
> Do you see any issues with the new lock file?

Yes, let's use pnpm v7, the same as Github actions.

PR #175, opened 2023-05-08
Improve error message

### LINKED ISSUE #174: Improve the error message when contract deployment fails
@nop33 reported that when contract deployment fails, the error message is something like this:

```
Loading alephium config file: /Users/ilias/dev/alephium/fetatoken/alephium.config.ts
Failed to deploy contracts, error: SyntaxError: Property assignment expected. (24:9)
  22 |     function toDeployments(json: any): Deployments {
  23 |       const contracts = {
> 24 |         ,
     |         ^
  25 |       }
  26 |       return {
  27 |         ...json,
```

We need to improve the error message.

Lbqds:
Resolve #174 

When the contract deployment fails, the deployed contract instance may be empty, and in this case, the generated code in deployments.ts cannot be compiled, resulting in strange error messages

### DIFF
--- a/packages/cli/scripts/deploy.ts
+++ b/packages/cli/scripts/deploy.ts
@@ -30,11 +30,11 @@ export async function deployAndSaveProgress<Settings = unknown>(
   try {
     await deploy(configuration, networkId, deployments)
   } catch (error) {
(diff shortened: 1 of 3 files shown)


polarker:
Nice fix 👍

PR #186, opened 2023-05-19
Allow defining multiple contracts/scripts in a single source file

Lbqds:
1. Allow defining multiple contracts/scripts in a single source file
2. Generate code for constants and enums


artifacts/ts/Assert.ts
```diff
@@ -24,14 +24,25 @@ import {
...
+  Error: bigint = BigInt(3);
+  A: bigint = BigInt(-3);
+  B: Address = "1DrDyTr9RpRsQnDnXo2YRiPzPW4ooHX5LLoqXrqfMrpQH";
+  C: HexString = "0011";
+  Addresses = {
+    A: "1DrDyTr9RpRsQnDnXo2YRiPzPW4ooHX5LLoqXrqfMrpQH",
+    B: "14UAjZ3qcmEVKdTo84Kwf4RprTQi86w2TefnnGFjov9xF",
+  };
+  Numbers = { A: BigInt(0), B: BigInt(1) };
+  ByteVecs = { A: "00", B: "11" };
```
polarker, line 44: Would be more friendly if we wrap all these contants in an object, like:
```
consts = {
  Error: ...,
  A: ...,
  ...
}

we could use Assert.consts.{Error, A, B, ...} instead of Assert..{Error, A, B, ...}?
```

polarker (commented):
LGTM, just one minor comment.

PR #188, opened 2023-05-23
Fix text endpoints

nop33:
@h0ngcha0, seems like the commit you added actually broke the parsing of the text endpoints response. This new PR fixes it (and also simplifies the code). All we had to do from the get-go was to simply add `"TEXT": '"text"'` in the `requestContentKind`, as you recommended.

Sorry that I didn't build the package locally after you added your commit to confirm that it still worked.


h0ngcha0 (approved):
LGTM

PR #190, opened 2023-05-26
Generate the getContractByCodeHash function

Lbqds:


artifacts/ts/contracts.ts
```diff
@@ -0,0 +1,42 @@
...
+  Assert,
+  Debug,
+  MetaData,
+  Warnings,
+  FakeTokenTest,
+  TokenTest,
+} from ".";
+
+export function getContractByCodeHash(codeHash: string): Contract {
+  const contracts: ContractFactory<any>[] = [
```
Lbqds, line 21: It doesn't work if move the `contracts` outside the function.

polarker: It should work if we use arrow function to capture the closure:
```
const contracts = ...

export const getContractByCodeHash = (codeHash: string): Contract => {
  ...
}
```

polarker (approved):
LGTM

PR #191, opened 2023-05-29
Fix webcrypto for chrome background service

Lbqds:
The `crypto.webcrypto` is undefined in chrome background service.


h0ngcha0:
I was trying to fix the same issue, thanks :+1:

h0ngcha0 (approved):
LGTM :+1:

PR #200, opened 2023-07-20
Fix the issue of could not find type declarations in vitejs

Lbqds:
I also tested other packages, the strange thing is that only `web3-react` has this issue.

### DIFF
--- a/packages/web3-react/package.json
+++ b/packages/web3-react/package.json
@@ -9,7 +9,12 @@
     "url": "git+https://github.com/alephium/alephium-web3-react"
   },
   "type": "module",
-  "exports": "./build/index.es.js",
+  "exports": {
+    ".": {
+      "types": "./build/index.d.ts",
+      "import": "./build/index.es.js"
+    }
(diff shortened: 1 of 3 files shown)


packages/web3-react/package.json
```diff
@@ -9,7 +9,12 @@
     "url": "git+https://github.com/alephium/alephium-web3-react"
   },
   "type": "module",
-  "exports": "./build/index.es.js",
+  "exports": {
+    ".": {
+      "types": "./build/index.d.ts",
+      "import": "./build/index.es.js"
```
Lbqds, line 15: Since we have declared `"type": "module"`, we only need to specify the `import` here.

packages/web3-react/package.json
```diff
@@ -9,7 +9,12 @@
...
+    }
+  },
   "types": "./build/index.d.ts",
```
Lbqds, line 18: We still need to keep this line, otherwise there will be the same error in reactjs.

PR #218, opened 2023-08-03
Improve useBalance hook

Lbqds:
I have tested it with DEX, and the balance can be updated as expected after tx confirmation.

### DIFF
--- a/packages/web3-react/src/hooks/useBalance.tsx
+++ b/packages/web3-react/src/hooks/useBalance.tsx
@@ -16,24 +16,49 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 import { Balance } from '@alephium/web3/dist/src/api/api-alephium'
-import { useEffect, useState } from 'react'
+import { useCallback, useEffect, useState } from 'react'
 import { useAlephiumConnectContext } from '../contexts/alephiumConnect'
+import { SubscribeOptions, Subscription, node, subscribeToTxStatus } from '@alephium/web3'
 
 export function useBalance() {
   const context = useAlephiumConnectContext()
   const [balance, setBalance] = useState<Balance>()
 
-  useEffect(() => {
-    const handler = async () => {
-      const nodeProvider = context.signerProvider?.nodeProvider
-      if (nodeProvider && context.account) {
-        const result = await nodeProvider.addresses.getAddressesAddressBalance(context.account.address)
-        setBalance(result)
-      }
+  const updateBalance = useCallback(async () => {
+    const nodeProvider = context.signerProvider?.nodeProvider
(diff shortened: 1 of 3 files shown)


packages/web3-react/src/hooks/useBalance.tsx
```diff
@@ -16,24 +16,49 @@ You should have received a copy of the GNU Lesser General Public License
...
     }
-
-    handler()
   }, [context.signerProvider?.nodeProvider, context.account])
 
-  return { balance }
+  const updateBalanceForTx = useCallback(
+    (txId: string, confirmations?: number) => {
+      const expectedConfirmations = confirmations ?? 1
+      const pollingInterval = context.network === 'devnet' ? 1000 : 10000
```
polarker, line 38: We may reduce the interval to 4 seconds (average block speed) since it's not very expensive

packages/web3-react/src/hooks/useBalance.tsx
```diff
@@ -16,24 +16,49 @@ You should have received a copy of the GNU Lesser General Public License
...
+        subscription.unsubscribe()
+        console.error(`tx status subscription error: ${err}`)
+        return Promise.resolve()
+      }
+      const options: SubscribeOptions<node.TxStatus> = {
+        pollingInterval,
+        messageCallback,
+        errorCallback
+      }
+      subscribeToTxStatus(options, txId, undefined, undefined, expectedConfirmations)
```
polarker, line 54: Does this require that we set up default node provider for web3?

packages/web3/src/contract/contract.ts
```diff
@@ -1786,3 +1788,20 @@ export async function getContractEventsCurrentCount(contractAddress: Address): P
       throw error
     })
 }
+
+export const getContractIdFromTxId = async (
```
h0ngcha0, line 1792: Questions:
1) What if we create multiple contracts in the same tx? 
2) if the created contract is a sub-contract, this doesn't apply, right?

packages/web3-react/src/hooks/useBalance.tsx
Lbqds: Yes, the `subscribeToTxStatus` function internally calls `web3.getCurrentNodeProvider()`.

packages/web3/src/contract/contract.ts
Lbqds: > What if we create multiple contracts in the same tx?

This function currently only supports creating one contract, not sure if we need to support multiple contracts

> if the created contract is a sub-contract, this doesn't apply, right?

Yes, the contract id of sub-contract is not calculated based on the tx output index

polarker: Let's add a comment to say this function only works in the simple case where a single non-subcontract is created in the tx.

polarker (approved):

PR #219, opened 2023-08-03
Generate wallets for testing

Lbqds:
The following is an example of a hardhat test:

```typescript
describe('test contract', () => {
  let wallet0: Wallet
  let wallet1: Wallet
  beforeAll(async () => {
    [wallet0, wallet1] = await getSigners()
  })
})
```

In hardhat, `getSigners` is used to load test wallets. When the user runs `hardhat test`, hardhat will start a devnet, and it will stop the devnet when the test is completed.

This is different from our usual practice, as we may keep the devnet running continuously, so it is still possible that ALPH has been exhausted, and users need to manually restart the devnet in this case.

### DIFF
--- a/packages/web3-test/src/index.ts
+++ b/packages/web3-test/src/index.ts
@@ -18,8 +18,17 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import 'cross-fetch/polyfill'
 
-import { addressFromContractId, isBase58, NodeProvider } from '@alephium/web3'
-import { NodeWallet } from '@alephium/web3-wallet'
+import {
+  addressFromContractId,
+  isBase58,
+  NodeProvider,
+  web3,
+  ONE_ALPH,
+  ALPH_TOKEN_ID,
+  DUST_AMOUNT,
+  Address
+} from '@alephium/web3'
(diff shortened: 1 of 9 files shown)


h0ngcha0 (approved):
LGTM :+1:

packages/web3-test/src/index.ts
```diff
@@ -67,6 +76,52 @@ export async function testNodeWallet(baseUrl = 'http://127.0.0.1:22973'): Promis
...
+  try {
+    return web3.getCurrentNodeProvider()
+  } catch (err) {
+    const nodeProvider = new NodeProvider('http://127.0.0.1:22973')
+    web3.setCurrentNodeProvider(nodeProvider)
+    return nodeProvider
+  }
+}
+
+export async function getSigner(alphAmount = ONE_ALPH * 100n): Promise<PrivateKeyWallet> {
```
polarker, line 89: We could make signer group an option `getSigner(alphAmount = ONE_ALPH * 100n, group = 0)`. The default group is `rootWallet.group = 0`.

polarker (approved):
LGTM, one minor remark

PR #222, opened 2023-08-05
Check contract/script initial fields

### LINKED ISSUE #220: Detect non hex string for fields
When providing name and symbol to deploy a new token, it's very easy to use string instead of hex string in the `initialFields`:
```
    initialFields: {
      symbol: 'TF', // Buffer.from('TF', 'utf8').toString('hex'),
      name: 'TokenFaucet' // Buffer.from('TokenFaucet', 'utf8').toString('hex'),
      decimals: 0n,
      supply: issueTokenAmount,
      balance: issueTokenAmount
    }
```

Lbqds:
Resolves #220

### DIFF
--- a/packages/web3/src/api/types.test.ts
+++ b/packages/web3/src/api/types.test.ts
@@ -1,0 +1,57 @@
+/*
+Copyright 2018 - 2022 The Alephium Authors
+This file is part of the alephium project.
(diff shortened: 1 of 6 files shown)


polarker (approved):
LGTM

PR #224, opened 2023-08-05
Fix the issue of re-rendering caused by useAccount

Lbqds:
Previously the return code for `useAccount` was: `return { account: context.account, isConnected: !!context.account }`. Once the wallet was connected, the `account` will not change.

Now the return code for `useAccount` is: `return context.account ? { networkType: context.network, ...context.account } : undefined`. And each time the `useAccount` is called, it returns a new object, which causes re-rendering.

### DIFF
--- a/packages/web3-react/src/hooks/useAccount.tsx
+++ b/packages/web3-react/src/hooks/useAccount.tsx
@@ -16,7 +16,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 import { getDefaultAlephiumWallet } from '@alephium/get-extension-wallet'
-import { useEffect } from 'react'
+import { useEffect, useMemo } from 'react'
 import { useAlephiumConnectContext } from '../contexts/alephiumConnect'
 import { KeyType } from '@alephium/web3'
 
@@ -56,5 +56,7 @@ export function useAccount(onDisconnected?: () => Promise<void>) {
     handler()
   }, [onDisconnected])
 
-  return context.account ? { networkType: context.network, ...context.account } : undefined
+  return useMemo(() => {
+    return context.account ? { networkType: context.network, ...context.account } : undefined
+  }, [context.account])
 }


packages/web3-react/src/hooks/useAccount.tsx
```diff
@@ -56,5 +56,7 @@ export function useAccount(onDisconnected?: () => Promise<void>) {
     handler()
   }, [onDisconnected])
 
-  return context.account ? { networkType: context.network, ...context.account } : undefined
+  return useMemo(() => {
+    return context.account ? { networkType: context.network, ...context.account } : undefined
+  }, [context.account])
```
polarker, line 61: How about this simpler approach:
```
const context = useAlephiumConnectContext()
const [account, setAccount] = useState<Account>()

useEffect(() => {...}

return account

Comment (Lbqds) [reply packages/web3-react/src/hooks/useAccount.tsx:61]: I tried the following two methods, but both of them does not work:

1.
```
useEffect(() => {
    const handler = async () => { ... }

    handler().then(() => setAccount(context.account))
  }, [onDisconnected, context])
```

2.
```
useEffect(() => {
    const handler = async () => {
      if (context.connectorId === 'walletConnect' || context.connectorId === 'desktopWallet') {
        return                                                         // <============== do we need to call setAccount here?
      }

      const keyType: KeyType = context.keyType ?? 'default'
      const connectedAccount = windowAlephium?.connectedAccount
      if (
        onDisconnected === undefined &&
        connectedAccount !== undefined &&
        connectedAccount.group === context.addressGroup &&
        connectedAccount.keyType === keyType &&
        windowAlephium?.connectedNetworkId === context.network
      ) {
        setAccount(connectedAccount)              // <============== setAccount
        return
      }

      const enabledAccount = await windowAlephium?.enableIfConnected({
        onDisconnected: onDisconnected ?? (() => Promise.resolve()),
        networkId: context.network,
        addressGroup: context.addressGroup,
        keyType: context.keyType
      })

      windowAlephium && context.setSignerProvider(windowAlephium)
      enabledAccount && context.setAccount(enabledAccount)
      setAccount(enabledAccount)                 // <============== setAccount
    }

    handler()
  }, [onDisconnected, context])
```

It only works if I add the second `useEffect`:

```
useEffect(() => {
  const handler() = async () => {... }
  handler()
}, [onDisconnected])

useEffect() => {
   setAccount(context.account)
}, [context.account])
```

But this should be equivalent to using `useMemo`, and I feel like we shouldn't put the auto-connect logic here because it only applies to the extension wallet.

polarker: This seems to work: https://github.com/alephium/alephium-web3/commit/0fc0e356a4b4918fac57627f1b0beded47b266ba

Let's check how `auto-connect` is done in `connect-kit`: https://github.com/family/connectkit

Lbqds: I tried it, but it still doesn't work, and there are two issues:

1. If the currently connected wallet is not the extension wallet, the return value of `useAccount` is always undefined.
2. Even if the currently connected wallet is an extension wallet, there are still two issues:

    a. After wallet connection, the connect button does not display as connected. This is because the `AlephiumConnectButton` also uses `useAccount`, but the `useAccount` handler is not triggered again after the wallet connection

    b. After disconnecting the wallet, the connect button still shows that the wallet is connected. The reason is the same as in issue `a`
    
And we cannot add `context.account` to the dependency list of the `useEffect` in `useAccount`. Otherwise, auto-connect will be triggered after the wallet is disconnected.

polarker (approved):
LGTM

PR #227, opened 2023-08-08
Refactor connect context

Lbqds:
Split the `AlephiumConnectContext` into two smaller contexts to avoid rerendering

### DIFF
--- a/packages/web3-react/src/components/AlephiumConnect.tsx
+++ b/packages/web3-react/src/components/AlephiumConnect.tsx
@@ -115,3 +115,28 @@ export const AlephiumConnectProvider: React.FC<{ children?: React.ReactNode }> =
 
   return <AlephiumConnectContext.Provider value={value}>{children}</AlephiumConnectContext.Provider>
 }
+
+export const AlephiumWalletProvider = ({
+  useTheme,
+  useMode,
+  useCustomTheme,
+  network,
+  addressGroup,
+  keyType,
+  children
+}: ConnectSettingProviderProps) => {
+  return (
+    <AlephiumConnectProvider>
+      <ConnectSettingProvider
+        useTheme={useTheme}
+        useMode={useMode}
+        useCustomTheme={useCustomTheme}
+        network={network}
+        addressGroup={addressGroup}
+        keyType={keyType}
+      >
+        {children}
+      </ConnectSettingProvider>
+    </AlephiumConnectProvider>
+  )
+}
--- a/packages/web3-react/src/components/AlephiumConnect.tsx
+++ b/packages/web3-react/src/components/AlephiumConnect.tsx
@@ -24,19 +24,19 @@ import { ThemeProvider } from 'styled-components'
 import { Account, KeyType, NetworkId, SignerProvider } from '@alephium/web3'
 import { Theme, Mode, CustomTheme } from '../types'
 import { routes } from './Common/Modal'
-import { AlephiumConnectContext, AlephiumConnectContextValue } from '../contexts/alephiumConnect'
+import { AlephiumConnectContext, ConnectSettingContext, ConnectSettingValue } from '../contexts/alephiumConnect'
 
-type AlephiumConnectProviderProps = {
+type ConnectSettingProviderProps = {
   useTheme?: Theme
   useMode?: Mode
   useCustomTheme?: CustomTheme
-  network?: NetworkId
-  addressGroup?: number
+  network: NetworkId
+  addressGroup: number
   keyType?: KeyType
   children?: React.ReactNode
 }
 
-export const AlephiumConnectProvider: React.FC<AlephiumConnectProviderProps> = ({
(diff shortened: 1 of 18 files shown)


packages/web3-react/src/hooks/useConnect.tsx
```diff
@@ -96,27 +96,29 @@ export function useConnect(options: ConnectOptions) {
...
+  }
 
   const connectAlephium = useCallback(async () => {
     const windowAlephium = await getDefaultAlephiumWallet()
 
     const enabledAccount = await windowAlephium
       ?.enable({
         ...options,
-        onDisconnected: disconnectAlephium
+        onDisconnected: () => {
```
h0ngcha0: Question: What's the reason to set `onDisconnected` to no-op?

Lbqds: Because we have already called `context.setSignerProvider(undefined)` and `context.setAccount(undefined)` at here: https://github.com/alephium/alephium-web3/blob/7cdf1fde136b81a459564c1c4b0285af86c4b96a/packages/web3-react/src/hooks/useConnect.tsx#L103

When the user clicks the disconnect button, the function call sequence is: `disconnectAlephium` -> `alephium.disconnect()` -> `onDisconnected`, which will cause `disconnectAlephium` to be executed twice. If I understand correctly, we only need to perform cleanup in either the `onDisconnect` callback or `disconnectAlephium`.

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -86,17 +80,38 @@ export const AlephiumConnectProvider: React.FC<AlephiumConnectProviderProps> = (
...
       <ThemeProvider theme={defaultTheme}>
         {children}
         <AlephiumConnectModal theme={theme} mode={mode} customTheme={customTheme} />
       </ThemeProvider>
-    </AlephiumConnectContext.Provider>
+    </ConnectSettingContext.Provider>
   )
 }
+
+export const AlephiumConnectProvider: React.FC<{ children?: React.ReactNode }> = ({ children }) => {
```
h0ngcha0, line 98: IIUC, now we need to wrap the app with both `AlephiumConnectProvider` and `ConnectSettingProvider`. Maybe we should update [README](https://github.com/alephium/alephium-web3/blob/master/packages/web3-react/README.md?plain=1#L38). I will update docs as well.

packages/web3-react/src/hooks/useConnect.tsx
Lbqds: Great point 👍, I just tested it, and the context will also be cleaned up in this case.

packages/web3-react/src/components/AlephiumConnect.tsx
h0ngcha0: Sorry, just saw in the next commit we have included `ConnectSettingProvider` in `AlephiumConnectProvider`.

Lbqds: Will update the readme, thanks!

h0ngcha0:
A general question:
Previously we have a single context, right now we split them into two and created `AlephiumWalletProvider` to wrap them together.  When will the previous case cause re-render while the new approach doesn't?

Lbqds:
> A general question: Previously we have a single context, right now we split them into two and created `AlephiumWalletProvider` to wrap them together. When will the previous case cause re-render while the new approach doesn't?

According to the [react doc](https://react.dev/reference/react/useContext#optimizing-re-renders-when-passing-objects-and-functions), when the value within the context updates, all components using the context will be rerendered. 

When we click the wallet connect button, both the `open` and `route` properties within the context will update, and then leading to multiple renders.

However, we have now split the context into two parts. Users only need to call `useAlephiumConnectContext` to retrieve wallet connection information, so updates to `ConnectSettingContext` will not trigger rerenders of the user's components. And his approach is inspired by `connectkit` and `wagmi`.

packages/web3/src/api/utils.ts
```diff
@@ -26,3 +27,20 @@ export function convertHttpResponse<T>(response: { status: number; data: T; erro
...
   }
 }
+
+export function isBalanceEqual(b0: node.Balance, b1: node.Balance): boolean {
+  const isTokenBalanceEqual = (tokens0?: node.Token[], tokens1?: node.Token[]): boolean => {
+    const tokens0Size = tokens0?.length ?? 0
+    const tokens1Size = tokens1?.length ?? 0
+    if (tokens0Size !== tokens1Size) return false
+    if (tokens0Size === 0) return true
+    return tokens0!.every((t0) => tokens1!.find((t1) => t0.id === t1.id && t0.amount === t1.amount) !== undefined)
```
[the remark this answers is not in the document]
Lbqds: Fixed, thanks!

polarker (approved):
LGTM

h0ngcha0 (approved):
LGMT :+1:

PR #234, opened 2023-08-09
Enable react hooks eslint plugin

Lbqds:
* Enable react hooks eslint plugin
* Add the balance to the connect context, and I have tested with ayin/dex/toolkit.


packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -105,8 +125,54 @@ export const AlephiumConnectProvider: React.FC<{ children?: React.ReactNode }> =
...
+        errorCallback
+      }
+      subscribeToTxStatus(options, txId, undefined, undefined, expectedConfirmations)
+    },
+    [updateBalance, account?.network]
+  )
+
+  useEffect(() => {
+    updateBalance()
+  }, [updateBalance])
```
polarker, line 171: This would need a balance query always, which is not necessary for all apps. Some apps might just want to get the account information.

Recommend:
1. Introduce another context for balance, which is `undefined` by default
2. `useBalance` would trigger the update the balance in the context if the balance is `undefined`, otherwise it will skip
3. `updateBalanceForTx` still works the same

Lbqds: Sounds great, will fix it, thanks!

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -163,23 +187,23 @@ export const AlephiumConnectProvider: React.FC<{ children?: React.ReactNode }> =
...
 
   useEffect(() => {
-    updateBalance()
-  }, [updateBalance])
+    if (balance === undefined) {
+      updateBalance()
+    } else if (connectContext.account === undefined) {
+      setBalance(undefined)
+    }
+  }, [updateBalance, balance, connectContext.account])
```
Lbqds, line 199: Although this won't lead to rerenders, the code here is not very good. The else branch is used to ensure that the balance is set to undefined when the wallet disconnects, or we could use a separate `useEffect`. WDYT?

packages/web3-react/src/components/Pages/Profile/index.tsx
```diff
@@ -18,23 +18,18 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
-import { AnimatePresence } from 'framer-motion'
-import { Balance, BalanceContainer, LoadingBalance } from './styles'
-import { prettifyAttoAlphAmount } from '@alephium/web3'
 import { useConnect } from '../../../hooks/useConnect'
 
 const Profile: React.FC<{ closeModal?: () => void }> = ({ closeModal }) => {
   const context = useConnectSettingContext()
   const account = useAccount()
-  const { balance } = useBalance()
   const { disconnect } = useConnect({
```
Lbqds, line 33: We have to remove `useBalance` from here, because `Profile` is not within the `AlephiumBalanceProvider` component tree.

polarker: We should add `AlephiumBalanceProvider` into `AlephiumWalletProvider`. Even if the app does not consume the provider, the overhead is insignificant.

polarker: But I agree that we should remove `useBalance` from `Profile`. Showing just ALPH is not that necessary

packages/web3-react/src/hooks/useBalance.tsx
```diff
@@ -15,11 +15,19 @@ GNU Lesser General Public License for more details.
...
 export function useBalance() {
   const context = useContext(AlephiumBalanceContext)
-  if (!context) throw Error('AlephiumBalance Hook must be inside a Provider.')
-  return context
+
+  useEffect(() => {
+    if (context?.balance === undefined) {
+      context?.updateBalance()
+    }
+  }, [context, context?.balance, context?.updateBalance])
```
polarker, line 28: no need to depend on `context`?

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -116,6 +142,69 @@ export const AlephiumConnectProvider: React.FC<{ children?: React.ReactNode }> =
...
+        }
+        return newBalance
+      })
+    } else if (account === undefined) {
+      setBalance(undefined)
+    }
+  }, [account, signerProvider])
+
+  const updateBalanceForTx = useCallback(
+    (txId: string, confirmations?: number) => {
```
polarker, line 170: Let's throw an error when `account === undefined`. One should not call `updateBalanceForTx` when account is not available

polarker (approved):
One minor comment left. Great work 👍 

@h0ngcha0 please take a look as well

packages/web3-react/src/components/ConnectModal/index.tsx
```diff
@@ -61,23 +61,19 @@ const ConnectModal: React.FC<{
...
+  }, [context])
 
   useEffect(() => {
-    if (isConnected) {
-      if (context.route !== routes.PROFILE) {
-        hide() // Hide on connect
-      }
-    } else {
-      hide() // Hide on connect
+    if (isConnected && context.route !== routes.PROFILE) {
```
h0ngcha0: Is it intentional to skip `hide()` when `isConnected` is false?

Lbqds: Yes, because the previous dependency list only included `isConnected`, it would only execute when `isConnected` changes. What we need is to hide this modal when the wallet is connected and the route is not `profile`, and keep it unchanged in other scenarios.

packages/web3-react/src/hooks/useBalance.tsx
```diff
@@ -15,11 +15,19 @@ GNU Lesser General Public License for more details.
...
+
+  return useMemo(() => {
```
h0ngcha0, line 30: [Nit] Do we need to wrap useMemo around `context?.balance`

packages/web3-react/src/components/ConnectModal/index.tsx
h0ngcha0: Got it, thanks for the explanation :+1:

h0ngcha0 (approved):
Awesome work, only minor questions :+1:

PR #245, opened 2023-08-24
Reset the state upon receiving the `session_delete` message

Lbqds:
When users click `disconnect` button in the desktop wallet, the dApp still shows that the wallet is connected. We need to reset the state upon receiving the `session_delete` message.


polarker (approved):
LGTM

PR #248, opened 2023-08-25
Add Standard Interface For NFT Collection With Royalty

h0ngcha0:


Lbqds (approved):
LGTM 👍

polarker:
@h0ngcha0 As Splinter pointed out on discord, most of the royalties design as NFT based, not collection based. That can be more flexible because one might mint multiple rounds for a same collection with different royalties.

h0ngcha0:
> @h0ngcha0 As Splinter pointed out on discord, most of the royalties design as NFT based, not collection based. That can be more flexible because one might mint multiple rounds for a same collection with different royalties.
> 
> We will need to update our NFT prototype to be NFT based loyalties.



> @h0ngcha0 As Splinter pointed out on discord, most of the royalties design as NFT based, not collection based. That can be more flexible because one might mint multiple rounds for a same collection with different royalties.
> 
> We will need to update our NFT prototype to be NFT based loyalties.

@polarker Royalty fee on Opensea is set at the collection level. (https://docs.opensea.io/docs/10-setting-fees-on-secondary-sales)

polarker:
> > @h0ngcha0 As Splinter pointed out on discord, most of the royalties design as NFT based, not collection based. That can be more flexible because one might mint multiple rounds for a same collection with different royalties.
> > We will need to update our NFT prototype to be NFT based loyalties.
> 
> @polarker
> 
> Question: Which NFT marketplace has NFT based royalty design? On Opensea Royalty fee is set at the collection level. (https://docs.opensea.io/docs/10-setting-fees-on-secondary-sales)
> 
> On EVM there is only the collection contract so I guess it has to be set at the collection level. However the royalty amount can be calculated per token id, that is why we have the following method in the interface
> 
> ```
> pub fn royaltyAmount(tokenId: ByteVec, salePrice: U256) -> (U256)
> ```

You are right. I missed that `tokenId` is enough for NFT-based royalties. All good then!

polarker (approved):
LGTM

PR #255, opened 2023-08-31
Fix web3 react

polarker:


packages/web3-react/src/components/ConnectButton/index.tsx
```diff
@@ -115,7 +114,6 @@ type ConnectButtonRendererProps = {
 }
 
 const ConnectButtonRenderer: React.FC<ConnectButtonRendererProps> = ({ displayAccount, children }) => {
-  const isMounted = useIsMounted()
```
polarker, line 118: This causes a rerender

Lbqds (approved):
LGTM

PR #257, opened 2023-08-31
web3-react more fixes

Lbqds:


packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -121,20 +117,76 @@ function tryGetNodeProvider(): NodeProvider | undefined {
...
+  const lastConnectedAccount = useMemo(() => {
+    const result = getLastConnectedAccount()
+    if (result === undefined) {
+      return undefined
+    }
+    if (
+      result.network === network &&
+      (addressGroup === undefined || result.account.group === addressGroup) &&
+      (keyType === undefined || result.account.keyType === keyType)
+    ) {
```
Lbqds, line 142: We need to check if the `addressGroup` or `network` has changed, so I moved them to the `AlephiumConnectContext`.

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -121,20 +117,76 @@ function tryGetNodeProvider(): NodeProvider | undefined {
...
+        setSignerProvider(undefined)
+      }
+      const onConnected = (result: ConnectResult) => {
+        updateAccount(result.account)
+        setSignerProvider(result.signerProvider)
+      }
+
+      try {
+        const result = await connector.autoConnect({ network, addressGroup, keyType, onDisconnected, onConnected })
+        if (result === undefined) {
```
Lbqds, line 172: When auto connect fails, we need to reset the state. This could happen when we reinstall the extension wallet.

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -121,20 +117,76 @@ function tryGetNodeProvider(): NodeProvider | undefined {
...
 
+  const updateAccount = useMemo(() => {
+    return (newAccount: Account | undefined) => {
+      setAccount((prev) => (prev?.address === newAccount?.address ? prev : newAccount))
+    }
+  }, [setAccount])
+
+  useEffect(() => {
+    const func = async () => {
+      if (lastConnectedAccount === undefined) return
```
Lbqds, line 158: Currently we auto connect based on `lastConnectedAccount.connectorId`, which means if `lastConnectedAccount` doesn't exist, we won't automatically reconnect even if we can. In wagmi it will retry the autoconnect for each connector if the last connector id does not exist. The code is here: https://github.com/wagmi-dev/wagmi/blob/4c614d524c7b0ac5b6817629468f5931815c1cd1/packages/core/src/config.ts#L243. Perhaps we can follow the same design, WDYT?

packages/web3-react/src/hooks/useWallet.tsx
```diff
@@ -20,22 +20,20 @@ import { useAlephiumConnectContext } from '../contexts/alephiumConnect'
...
-  account: Account & { network: NetworkId }
-  nodeProvider?: NodeProvider
+  signer: SignerProvider | undefined
+  account: (Account & { network: NetworkId }) | undefined
+  nodeProvider: NodeProvider | undefined
 }
 
 export function useWallet() {
-  const { account, signerProvider } = useAlephiumConnectContext()
+  const { account, signerProvider, network } = useAlephiumConnectContext()
```
Lbqds, line 29: When auto reconnecting, it is possible that the `account` is not undefined but the `signerProvider` is undefined, so I changed the return value of the `useWallet` hook.

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -121,20 +117,76 @@ function tryGetNodeProvider(): NodeProvider | undefined {
...
+      }
+
+      try {
+        const result = await connector.autoConnect({ network, addressGroup, keyType, onDisconnected, onConnected })
+        if (result === undefined) {
+          updateAccount(undefined)
+          removeLastConnectedAccount()
+        }
+      } catch (error) {
+        console.error(error)
```
polarker, line 177: Should we also remove the last connected account here?

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -43,8 +43,10 @@ import {
   ConnectSettingValue,
   useAlephiumConnectContext
 } from '../contexts/alephiumConnect'
+import { getLastConnectedAccount, removeLastConnectedAccount } from '../utils/storage'
+import { ConnectResult, getConnectorById } from '../utils/connector'
 
-type ConnectSettingProviderProps = {
+type AlephiumWalletProviderProps = {
```
polarker, line 49: Let's move this just above `AlephiumWalletProvider`

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -121,20 +117,76 @@ function tryGetNodeProvider(): NodeProvider | undefined {
...
+    if (result === undefined) {
+      return undefined
+    }
+    if (
+      result.network === network &&
+      (addressGroup === undefined || result.account.group === addressGroup) &&
+      (keyType === undefined || result.account.keyType === keyType)
+    ) {
+      return result
+    }
```
polarker, line 144: Do we need to return `undefined` by default?

packages/web3-react/src/components/AlephiumConnect.tsx
polarker: Yes, we could do the same.

packages/web3-react/src/components/AlephiumConnect.tsx
polarker: Makes more sense to be in `AlephiumConnectContext`

packages/web3-react/src/components/AlephiumConnect.tsx
Lbqds: Great catch, will fix it, thanks!

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -121,20 +117,76 @@ function tryGetNodeProvider(): NodeProvider | undefined {
...
+  }, [network, addressGroup, keyType])
+
+  const [account, setAccount] = useState<Account | undefined>(lastConnectedAccount?.account)
   const [signerProvider, setSignerProvider] = useState<SignerProvider | undefined>()
 
+  const updateAccount = useMemo(() => {
```
h0ngcha0, line 166: Q: Should this be `useCallback`?

packages/web3-react/src/hooks/useWallet.tsx
```diff
@@ -16,25 +16,52 @@ You should have received a copy of the GNU Lesser General Public License
...
+      signer: SignerProvider
+      account: Account & { network: NetworkId }
+      connectionStatus: 'connected'
+      nodeProvider: NodeProvider | undefined
+    }
+  | {
+      signer: SignerProvider | undefined
+      account: (Account & { network: NetworkId }) | undefined
+      connectionStatus: 'connecting' | 'reconnecting'
+      nodeProvider: NodeProvider | undefined
```
polarker, line 33: both `signer` and `nodeProvider` should be `undefined` when the wallet is connecting?

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -148,18 +171,21 @@ export const AlephiumConnectProvider: React.FC<{
 
   useEffect(() => {
     const func = async () => {
+      if (connectionStatusRef.current !== 'connecting') return
+
       const onDisconnected = () => {
```
h0ngcha0, line 176: Q: If a connector doesn't have `autoConnect` defined, e.g. `WalletConnect`, and the `signer` was `undefined` before. Is it possible that `connectionStatus` is stuck at `connecting`?

packages/web3-react/src/components/AlephiumConnect.tsx
Lbqds: Yeah, `useCallback` is better than `useMemo`, fixed.

packages/web3-react/src/components/AlephiumConnect.tsx
Lbqds: No, if auto connect fails, it will call the `onDisconnected` callback, which set the status to `disconnected`.

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -138,41 +138,27 @@ export const AlephiumConnectProvider: React.FC<{
     return undefined
   }, [network, addressGroup, keyType])
 
-  const connectionStatusRef = useRef<ConnectionStatus>('connecting')
+  const [connectionStatus, setConnectionStatus] = useState<ConnectionStatus>('connecting')
```
Lbqds, line 141: `useRef` does not rerender when connection status changes, so I replaced it with `useState` here. This means that there will be a rerender when the connection status changes.

If the initial status is `disconnected`, in the `useWallet` hook, the account will be undefined, so I set the initial status to `connecting`.

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -208,14 +194,15 @@ export const AlephiumConnectProvider: React.FC<{
     }
 
     func()
-  }, [lastConnectedAccount, network, addressGroup, keyType, updateSignerProvider, updateAccount, connectionStatusRef])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
Lbqds, line 198: We should only need to auto connect once when the page refreshes, so I've set this dependency list to empty, and it also auto connect once in wagmi.

packages/web3-react/src/components/AlephiumConnect.tsx
polarker: We could initialize it like this:
```
  const [connectionStatus, setConnectionStatus] = useState<ConnectionStatus>(
    lastConnectedAccount ? 'connecting' : 'disconnected'
  )
```

packages/web3-react/src/hooks/useWallet.tsx
```diff
@@ -19,23 +19,49 @@ import { useMemo } from 'react'
...
-  account: Account & { network: NetworkId }
-  nodeProvider?: NodeProvider
-}
+export type Wallet =
+  | {
+      signer: SignerProvider
+      account: Account & { network: NetworkId }
+      connectionStatus: 'connected'
+      nodeProvider: NodeProvider | undefined
+    }
```
polarker, line 28: Recommend placing `connectionStatus` at the first field so new devs knows that it is the type field.
```
    connectionStatus: 'connected';
    signer: SignerProvider;
    account: Account & {
        network: NetworkId;
    };
    nodeProvider: NodeProvider | undefined;
```

packages/web3-react/src/components/AlephiumConnect.tsx
Lbqds: In the latest code here: https://github.com/alephium/alephium-web3/pull/257/files#diff-d9969eee0099bfba3bd0dd6f677ddfc279b75331f2cf79de5f1545bac4315cb9R150, it sets the `connectionStatus` to `disconnected` if the `newSignerProvider` is undefined.

polarker (approved):
LGTM 👍

PR #258, opened 2023-09-01
Generate default fields to facilitate deployment of template contracts

Lbqds:


h0ngcha0 (approved):
LGTM :+1:

artifacts/ts/Greeter.ts
```diff
@@ -56,6 +64,39 @@ export namespace GreeterTypes {
...
+      [
+        "tgx7VNFoP9DJiFMFgXXtafQZkUvyEdDHT9ryamHJYrjq",
+        "tgx7VNFoP9DJiFMFgXXtafQZkUvyEdDHT9ryamHJYrjq",
+      ],
+      [
+        "tgx7VNFoP9DJiFMFgXXtafQZkUvyEdDHT9ryamHJYrjq",
+        "tgx7VNFoP9DJiFMFgXXtafQZkUvyEdDHT9ryamHJYrjq",
+      ],
+    ],
+  };
```
polarker, line 98: This would generate a lot of code. Is it possible to do it this way:
```
function getDefaultInitialFields(): GreeterTypes.Fields = {
  // call a shared function from contract.ts
}
```

packages/web3-test/src/token.ts
```diff
@@ -0,0 +1,92 @@
...
+        address
+      )
+    }
+    ${contractCode}
+  `
+  const scriptResult = await nodeProvider.contracts.postContractsCompileScript({ code: scriptCode })
+  return Script.fromCompileResult(scriptResult)
+}
+
+export async function createAndTransferToken(nodeProvider: NodeProvider, signer: SignerProvider, amount: bigint) {
```
polarker, line 84: 1. We should use the default addresses to deploy the script and execute the script, so that the balance of recipient is not changed. Otherwise it's hard to test the balance changes
2. signer should be renamed as `recipient: Address`

packages/web3-test/src/index.ts
```diff
@@ -141,6 +143,13 @@ export async function transfer(from: PrivateKeyWallet, to: Address, tokenId: str
   return await from.signAndSubmitTransferTx({ signerAddress: from.address, destinations: [destination] })
 }
 
+export async function mintToken(from: PrivateKeyWallet, amount: bigint) {
+  const nodeProvider = tryGetDevnetNodeProvider()
+  const result = await createAndTransferToken(nodeProvider, from, amount)
+  const contractId = await getContractIdFromUnsignedTx(nodeProvider, result.unsignedTx)
+  return { ...result, contractId, contractAddress: addressFromContractId(contractId) }
+}
```
polarker, line 151: let's create `const.ts` and then move this function to `token.ts`. We don't need to expose `createAndTransferToken` I think.

packages/web3-test/src/token.ts
Lbqds: Currently we only have a test address on group 0, if we want to use the default address to deploy the script, do we need to export test addresses from other groups? Because the user might want to deploy the token on other groups, or we can deploy the contract on group 0 and then transfer token, but this need two transactions, WDYT?

polarker: Yes, we should include a default address for each group, and only expose the group 0 one for external usage.

packages/web3-test/src/index.ts
Lbqds: I added an index.ts file for exporting what we want to export.

packages/web3-test/src/token.ts
Lbqds: > 1. We should use the default addresses to deploy the script and execute the script, so that the balance of recipient is not changed. Otherwise it's hard to test the balance changes

The balance still changes when minting tokens, we need to transfer the dust amount to the recipient here: https://github.com/alephium/alephium-web3/pull/258/commits/2d8542013c222e0a373080d07ca44e937b85fbab#diff-533abf520a4fdc1637791a205da159e5ce4c886c63fe8d00505e5961d47fd82aR84. Currently we do not support constructing transactions with inputs from different addresses.

test/contract.test.ts
```diff
@@ -165,12 +165,13 @@ describe('contract', function () {
   })
 
   it('should test contract (2)', async () => {
-    const testResult = await Greeter.tests.greet({ initialFields: { ...Greeter.defaultInitialFields, btcPrice: 1n } })
+    const initialFields = Greeter.getDefaultInitialFields()
```
polarker, line 168: The naming is a bit misleading. How about `genInitialFieldsWithDefaultValues()`?

packages/web3-test/src/const.ts
```diff
@@ -0,0 +1,43 @@
...
+export const testMnemonic =
+  'vault alarm sad mass witness property virus style good flower rice alpha viable evidence run glare pretty scout evil judge enroll refuse another lava'
+export const testWalletName = 'alephium-web3-test-only-wallet'
+export const testAddress = '1DrDyTr9RpRsQnDnXo2YRiPzPW4ooHX5LLoqXrqfMrpQH'
+export const testPrivateKey = testPrivateKeys[0]
+export const testPassword = 'alph'
+
+export function tryGetDevnetNodeProvider(): NodeProvider {
+  try {
+    return web3.getCurrentNodeProvider()
```
polarker, line 37: Let's check whether the network is devnet as well. New devs might setTestnet sometimes

polarker (approved):
LGTM

PR #259, opened 2023-09-01
Simplify theme

polarker:


packages/web3-react/src/types.ts
```diff
@@ -17,7 +17,18 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+export type ProviderTheme =
+  | 'simple-light'
+  | 'simple-dark'
+  | 'web95'
+  | 'retro'
+  | 'soft'
+  | 'midnight'
+  | 'minimal'
+  | 'rounded'
+  | 'nouns'
```
polarker, line 29: I found this is more friendly as only one theme has light and dark mode

h0ngcha0 (approved):
LGTM

PR #261, opened 2023-09-01
Improve web3 react

polarker:
* Hide all of the contexts in favour of hooks
* Export configuration via hook

### DIFF
--- a/packages/web3-react/src/hooks/useWallet.tsx
+++ b/packages/web3-react/src/hooks/useWallet.tsx
@@ -18,6 +18,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 import { useMemo } from 'react'
 import { useAlephiumConnectContext } from '../contexts/alephiumConnect'
 import { NodeProvider, SignerProvider, Account, NetworkId, ExplorerProvider } from '@alephium/web3'
+import { KeyType } from '@alephium/web3'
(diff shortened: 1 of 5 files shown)


packages/web3-react/src/hooks/useWallet.tsx
```diff
@@ -71,3 +72,27 @@ export function useWallet() {
...
+  network: NetworkId
+  setNetwork: (network: NetworkId) => void
+  addressGroup?: number
+  setAddressGroup: (addressGroup: number | undefined) => void
+  keyType: KeyType
+  setKeyType: (keyType: KeyType) => void
+}
+
+export function useWalletConfig() {
+  const { network, setNetwork, addressGroup, setAddressGroup, keyType, setKeyType } = useAlephiumConnectContext()
```
Lbqds, line 86: Question: We disconnect when the `network` changes for now. Do we also want to disconnect when the `addressGroup` and `keyType` change?

Lbqds (approved):
only one question, otherwise LGTM

packages/web3-react/src/hooks/useWallet.tsx
polarker: Great catch. Just added the fix.

Lbqds (approved):

PR #263, opened 2023-09-04
Add withdraw method to INFTCollectionWithRoyalty

h0ngcha0:


packages/web3/std/nft_collection_with_royalty_interface.ral
```diff
@@ -4,4 +4,6 @@ Interface INFTCollectionWithRoyalty extends INFTCollection {
 
     @using(preapprovedAssets = true)
     pub fn payRoyalty(payer: Address, amount: U256) -> ()
+
+    pub fn withdraw(to: Address, amount: U256) -> ()
```
polarker, line 8: How about `withdrawRoylty`?

h0ngcha0: Sounds good, thanks :+1:

PR #264, opened 2023-09-05
Minor fixes

polarker:


packages/web3-react/src/utils/storage.ts
```diff
@@ -28,7 +28,7 @@ export type ConnectedAccount = {
 }
 
 function getStorage() {
-  return window === undefined ? undefined : window.localStorage
+  return globalThis.window?.localStorage
```
polarker, line 31: This would work in non-browser environment

Lbqds (approved):
LGTM

h0ngcha0 (approved):
LGTM :+1:

PR #265, opened 2023-09-07
Update NFT Interfaces

h0ngcha0:
1. Add `getCollectionIndex` to  `INFT` interface
2. Add `validateNFT` to `INFTCollection` interface


packages/web3/std/nft_collection_interface.ral
```diff
@@ -26,4 +26,6 @@ Interface INFTCollection {
    pub fn totalSupply() -> U256
 
    pub fn nftByIndex(index: U256) -> INFT
+
+   pub fn validateNFT(nftId: ByteVec, nftIndex: U256) -> ()
```
polarker, line 30: Let's add a comment that this should throw an error if the validation failed.

h0ngcha0: Sounds great :+1:

h0ngcha0:
NFT codebase now uses this two extra methods now. Here is the [PR](https://github.com/alephium/alephium-nft/pull/36/files) for it.

polarker:
@h0ngcha0 It seems you will need this PR in the NFT repo now. Let's merge it and release `0.22.0`?

h0ngcha0:
> @h0ngcha0 It seems you will need this PR in the NFT repo now. Let's merge it and release `0.22.0`?

Yeah, one last thing. The `collectionId` that we just added to the `NFTMetadata` is a breaking change for `nodeProvider.fetchNFTMetaData`, since some of the NFTs do not have `getCollectionIndex` or `getCollectionId` method defined. Shall we make `collectionId` optional for `NFTMetadata` instead?

polarker:
> Yeah, one last thing.

That's a tricky question.
1. We should add `nftIndex` to the metadata as well.
2. I feel that it's better to just introduce the break change, anyway the marketplace would not be able to support the old NFTs now. If getNftMetadata fails due to the `getCollectionIndex` function, then we treat it as deprecated NFT and don't support them anymore in the new marketplace.

h0ngcha0:
> > Yeah, one last thing.
> 
> That's a tricky question.
> 
> 1. We should add `nftIndex` to the metadata as well.
> 2. I feel that it's better to just introduce the break change, anyway the marketplace would not be able to support the old NFTs now. If getNftMetadata fails due to the `getCollectionIndex` function, then we treat it as deprecated NFT and don't support them anymore in the new marketplace.
> 
> The early we introduce the break, the better.

It's not only the marketplace, wallets and explorer might also have issues displaying some of the NFTs due to this breaking change. If we make `nftIndex` mandatory, we might have issues displaying all existing NFTs on wallet and explorer if we use `nodeProvider.fetchNFTMetaData`.

I agree it is better to have breaking change now rather than later though.

packages/web3/src/api/node-provider.ts
```diff
@@ -133,10 +133,20 @@ export class NodeProvider implements NodeProviderApis {
...
+    const tokenUri = hexToString(tryGetCallResult(result.results[0]).returns[0].value as any as string)
+    const collectionIndexResult = result.results[1]
+    if (collectionIndexResult.type === 'CallContractSucceeded') {
+      const successfulCollectionIndexResult = result.results[1] as CallContractSucceeded
+      const collectionId = successfulCollectionIndexResult.returns[0].value as any as string
+      const nftIndex = BigInt(successfulCollectionIndexResult.returns[1].value as any as string)
+      return { tokenUri, collectionId, nftIndex }
+    } else {
+      const failedCollectionIndexResult = result.results[1] as CallContractFailed
+      if (failedCollectionIndexResult.error.startsWith('VM execution error: InvalidMethodIndex')) {
```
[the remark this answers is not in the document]
h0ngcha0: Great catch, previous standard had `getCollectionId` method. Updated now.

contracts/nft/deprecated_nft_test_2.ral
```diff
@@ -0,0 +1,12 @@
+Contract DeprecatedNFTTest2(
+  collectionId: ByteVec,
+  uri: ByteVec
+) {
+  pub fn getTokenUri() -> ByteVec {
+    return uri
+  }
+
+  pub fn getCollectionId() -> ByteVec {
```
polarker, line 9: What would happen if the return type is boolean?

h0ngcha0: Added more tests now, thanks!

contracts/nft/deprecated_nft/deprecated_nft_test_4.ral
```diff
@@ -0,0 +1,12 @@
+Contract DeprecatedNFTTest4(
```
polarker, line 1: We should be able to put all these test in a same file now.

polarker (approved):
LGTM, one small remark.

contracts/nft/deprecated_nft_test_2.ral
polarker: LGTM.

PR #270, opened 2023-09-11
[WIP]Walletconnect: auto connect

Lbqds:
Need to do more testing, the branch for the desktop wallet is here: https://github.com/alephium/desktop-wallet/tree/walletconnect-auto-connect


packages/walletconnect/src/provider.ts
```diff
@@ -113,26 +113,47 @@ export class WalletConnectProvider extends SignerProvider {
...
       }
-    })
 
-    if (uri) {
-      this.emitEvents('displayUri', uri)
+      this.session = await approval()
+      this.updateNamespace(this.session.namespaces)
+    } else {
+      this.updateNamespace(this.session.namespaces)
+      await this.ping(this.session.topic)
```
Lbqds, line 135: When the dApp auto connects but the wallet is not open, sending requests will be blocked. Since walletconnect allows a timeout range of 5 minutes to 7 days, we need to set an appropriate timeout ourselves.

polarker: We should add a timeout to extension wallet connection as well, so the wallet will not auto-connect forever

packages/web3-react/src/utils/connector.ts
```diff
@@ -142,7 +143,13 @@ const injectedDisconnect = async (signerProvider: SignerProvider): Promise<void>
...
   try {
     const windowAlephium = await getDefaultAlephiumWallet()
-    const enabledAccount = await windowAlephium?.enableIfConnected({ ...options, networkId: options.network })
+    if (windowAlephium === undefined) {
+      throw new Error('Extension wallet not found')
+    }
+    const enabledAccount = await timeout(
+      windowAlephium.enableIfConnected({ ...options, networkId: options.network }),
+      AUTO_CONNECT_TIMEOUT
+    )
```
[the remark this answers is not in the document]
Lbqds: Sorry I misunderstood, this sounds good, I will revert the last commit and add an expiry to the extended wallet.

polarker (approved):
LGTM

h0ngcha0 (approved):
LGTM :+1:

PR #271, opened 2023-09-14
Support getting contract deployment result from other groups

Lbqds:
Sometimes we need to retrieve contract results deployed on other groups, such as in the ANS contract, where the `PrimaryRegistrar` contract is deployed on group 0, and the `SecondaryRegistrar` deployed on other groups depends on the contract id of `PrimaryRegistrar`.


polarker (approved):
LGTM

PR #272, opened 2023-09-14
Reorder fields of MetaData

tdroxler:
In order to respect the extract order defined
in https://github.com/alephium/alephium-web3/blob/9c84d2926f5e052493bb9d0be8af48ffe1efd432/packages/web3/src/api/node-provider.ts


h0ngcha0 (approved):
LGTM

PR #276, opened 2023-11-11
Tx Codec

h0ngcha0:


packages/web3/src/codec/transaction-codec.ts
```diff
@@ -0,0 +1,110 @@
...
+export const signaturesCodec = new ArrayCodec(signatureCodec)
+const contractOutputRefsCodec = new ArrayCodec(contractOutputRefCodec)
+
+export interface Transaction {
+  unsigned: UnsignedTransaction
+  scriptExecutionOk: number
+  contractInputs: DecodedArray<ContractOutputRef>
+  generatedOutputs: DecodedArray<Output>
+  inputSignatures: DecodedArray<Signature>
+  scriptSignatures: DecodedArray<Signature>
```
Lbqds, line 42: We need to call the `convertTo*` function after decoding, is it possible to directly declare the type to `ContractOutputRef[]`, `Output[]`, etc?

Or we can add some util functions like this:

```
function decodeTransaction(bytes: Uint8Array): Transaction {
  const tx = transactionCodec.decode(bytes)
  return TransactionCodec.convertTo(tx)
}
```

h0ngcha0: Sorry for the late reply, will do :+1:

packages/web3/src/index.ts
```diff
@@ -29,5 +29,5 @@ export * from './token'
 
 export * from './constants'
 export * as web3 from './global'
-
+export * as codec from './codec'
```
polarker, line 32: As discussed, we don't want to expose new functions via root `web3/src/index.ts`.

h0ngcha0: Got it :+1:

packages/web3/src/codec/codec.ts
```diff
@@ -0,0 +1,30 @@
...
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+import { Parser } from 'binary-parser'
+
+export interface Codec<T> {
+  parser: Parser
+  encode(input: T): Buffer
```
Lbqds, line 22: It seems there are some issues with the `Buffer`, it's not compatible with the buffer we are using in `web3` package. We need to import it from `buffer/`: `import { Buffer } from `buffer/`, I have fixed it in the `ralph-map` branch: https://github.com/alephium/alephium-web3/pull/337/commits/4bc2c24aca06cc541c1993852b0105e083ee0d3d

packages/web3/src/codec/asset-output-codec.ts
```diff
@@ -0,0 +1,150 @@
...
+      hint = createHint((outputLockupScript as PublicKeyHash).publicKeyHash)
+    } else if (scriptType === 1) {
+      // P2MPKH
+      hint = createHint((outputLockupScript as MultiSig).publicKeyHashes.value[0].publicKeyHash)
+    } else if (scriptType === 2) {
+      // P2SH
+      hint = createHint((outputLockupScript as P2SH).scriptHash)
+    } else if (scriptType === 3) {
+      // P2C
+      hint = createHint((outputLockupScript as P2C).contractId)
```
Lbqds, line 103: [Nit]: Asset output cannot be P2C, maybe we can throw an error here.

packages/web3/src/codec/asset-output-codec.ts
```diff
@@ -0,0 +1,150 @@
...
+    } else {
+      throw new Error(`TODO: decode output script type: ${scriptType}`)
```
Lbqds, line 105: [Nit]: I guess we can remove the `TODO` now.

packages/web3/src/codec/asset-output-codec.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  static fromFixedAssetOutputs(fixedOutputs: FixedAssetOutput[]): AssetOutput[] {
+    return fixedOutputs.map((output) => {
+      return AssetOutputCodec.fromFixedAssetOutput(output)
+    })
+  }
+
+  static fromFixedAssetOutput(fixedOutput: FixedAssetOutput): AssetOutput {
+    const amount: DecodedCompactInt = compactUnsignedIntCodec.decode(
+      compactUnsignedIntCodec.encodeU256(BigInt(fixedOutput.attoAlphAmount))
```
Lbqds, line 119: After encoding, we decoded once again. Is it possible to directly convert bigint to `DecodedCompactInt`?

packages/web3/src/codec/contract-output-codec.ts
```diff
@@ -0,0 +1,97 @@
...
+    .nest('lockupScript', {
+      type: Parser.start().buffer('contractId', { length: 32 })
+    })
+    .nest('tokens', {
+      type: tokensCodec.parser
+    })
+
+  encode(input: ContractOutput): Buffer {
+    const amount = Buffer.from(compactUnsignedIntCodec.encode(input.amount))
+    const lockupScript = (input.lockupScript as P2C).contractId
```
Lbqds, line 51: [Nit]: `as P2C` is not necessary.

packages/web3/src/codec/instr-codec.ts
```diff
@@ -0,0 +1,462 @@
...
+export const CreateSubContractAndTransferToken: Instr = { code: 0xc8, value: {} }
+export const CopyCreateSubContractAndTransferToken: Instr = { code: 0xc9, value: {} }
+export const NullContractAddress: Instr = { code: 0xca, value: {} }
+export const SubContractId: Instr = { code: 0xcb, value: {} }
+export const SubContractIdOf: Instr = { code: 0xcc, value: {} }
+export const AlphTokenId: Instr = { code: 0xcd, value: {} }
+export const LoadImmField = (index: number): Instr => ({ code: 0xce, value: { index } })
+export const LoadImmFieldByIndex: Instr = { code: 0xcf, value: {} }
+
+export class InstrCodec implements Codec<Instr> {
```
Lbqds, line 239: We added some new instructions in the Rhone upgrade. Do we need to support them now?

.github/workflows/test.yml
```diff
@@ -19,9 +19,9 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
-      - uses: pnpm/action-setup@v2.2.4
+      - uses: pnpm/action-setup@v3.0.0
         with:
-          version: 7.19
+          version: 8.14
```
Lbqds, line 24: It seems that pnpm 7.19 also works. Just curious, what is the reason for upgrading the pnpm version?

h0ngcha0: IIRC, these two versions generate different `pnpm.lock` file, which will fail CI since CI checks if the git repo is clean after building

packages/web3/src/codec/instr-codec.ts
h0ngcha0: I will add, since they it wouldn't break anything here.

packages/web3/src/codec/asset-output-codec.ts
h0ngcha0: `compactUnsignedIntCodec.decode` should be pretty efficient since it just takes the first byte, and store rest of the bytes into a `UintArray`. I can create a method to do `compactUnsignedIntCodec.decode(BigInt(fixedOutput.attoAlphAmount))`, WDYT?

.github/workflows/test.yml
Lbqds: Thanks for the explanation, I'll update the pnpm version

Lbqds (approved):
LGTM 👍

PR #277, opened 2023-12-02
Add some util functions for the exchange

Lbqds:


packages/web3/src/utils/exchange.ts
```diff
@@ -0,0 +1,120 @@
...
+  if (decoded.length === 0) throw new Error('UnlockScript is empty')
+  const unlockScriptType = decoded[0]
+  const unlockScriptBody = decoded.slice(1)
+
+  if (unlockScriptType === UnlockScriptType.P2PKH) {
+    return addressFromPublicKey(binToHex(unlockScriptBody))
+  } else if (unlockScriptType === UnlockScriptType.P2MPKH) {
+    throw new Error('Multisig address is not supported')
+  } else if (unlockScriptType === UnlockScriptType.P2SH) {
+    // FIXEME: for now we assume that the params is empty, so we need to
```
Lbqds, line 40: Because the unlock script consists of both script and params, we need to use the codec in this PR: https://github.com/alephium/alephium-web3/pull/276. For now we only have the schnorr script, and its params are empty.

test/exchange.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    })
+    await waitTxConfirmed(web3.getCurrentNodeProvider(), result.txId, 1, 1000)
+    this.depositTxs.push({ txId: result.txId, from: from.address })
+    const accountBalance = this.accountBalances.get(from.address)
+    if (accountBalance === undefined) {
+      this.accountBalances.set(from.address, amount)
+    } else {
+      this.accountBalances.set(from.address, accountBalance + amount)
+    }
+  }
```
polarker, line 60: The `Exchange` should poll new blocks and find deposit transactions with the helper functions from `exchange.ts`. Maybe we should have another class `User` which only deposit ALPH to an exchange.

packages/web3/src/utils/exchange.ts
```diff
@@ -0,0 +1,120 @@
...
+// we assume that the tx is deposit transaction
+export function getDepositAddress(tx: Transaction): Address {
+  return getAddressFromUnlockScript(tx.unsigned.inputs[0].unlockScript)
+}
+
+export function isExchangeAddress(address: string): boolean {
+  const decoded = bs58.decode(address)
+  if (decoded.length === 0) throw new Error('Address is empty')
+  const addressType = decoded[0]
+  return addressType === AddressType.P2PKH || addressType === AddressType.P2SH
```
polarker, line 46: let's also validate the length of the hash

packages/web3/src/utils/exchange.ts
```diff
@@ -0,0 +1,120 @@
...
+export function isDepositTokenTransaction(tx: Transaction, exchangeAddress: string): boolean {
+  return isDepositTransaction(tx, exchangeAddress) && checkTokenOutput(tx, exchangeAddress)
+}
+
+// we assume that the tx is deposit transaction
+export function getDepositAddress(tx: Transaction): Address {
+  return getAddressFromUnlockScript(tx.unsigned.inputs[0].unlockScript)
+}
+
+export function isExchangeAddress(address: string): boolean {
```
polarker, line 42: Let's put this one at the beginning

packages/web3/src/utils/exchange.ts
```diff
@@ -0,0 +1,120 @@
...
+  const decoded = bs58.decode(address)
+  if (decoded.length === 0) throw new Error('Address is empty')
+  const addressType = decoded[0]
+  return addressType === AddressType.P2PKH || addressType === AddressType.P2SH
+}
+
+export function getAddressFromUnlockScript(unlockScript: string): Address {
```
polarker, line 49: Do we want to expose this?

Lbqds: We need to expose it because we are using it in unit tests, but now in `index.ts` it is not exposed.

test/exchange.test.ts
```diff
@@ -116,47 +165,58 @@ describe('exchange', function () {
...
+    for (const user of users) {
+      const depositAmount = user.getDepositAmount()
+      totalDepositAmount += user.getDepositAmount()
+      const txsByUser = user.getDepositTxs()
+      const gasFee = await getGasFee(txsByUser)
+      const balance = await nodeProvider.addresses.getAddressesAddressBalance(user.address)
+      expect(BigInt(balance.balance)).toEqual(initialBalance - depositAmount - gasFee)
     }
+    const exchangeBalance = await nodeProvider.addresses.getAddressesAddressBalance(exchangeWallet.address)
+    expect(BigInt(exchangeBalance.balance)).toEqual(initialBalance + totalDepositAmount)
```
polarker, line 220: Do we need to unsubscribe polling at the end?

polarker (approved):
LGTM

PR #282, opened 2023-12-08
Redirect to deep link

Lbqds:
I have tested the desktop wallet, android wallet, and ios wallet using this branch. It can successfully redirect from browser to wallet app. But there are still some issues:

1. The current desktop wallet automatically pops up when it receives a request, and the browser also displays a request to  redirect to the desktop wallet. I think we can solve this by:

      a. Delete this line of code in the desktop wallet: https://github.com/alephium/alephium-frontend/blob/4ff8fd822ed998d6d2e8745c3ddeb2f176a298e2/apps/desktop-wallet/src/contexts/walletconnect.tsx#L186
      b. Or in the web3 sdk, we only redirect to wallet if it's mobile
  
2. When I used the **_ios simulator_** to test, I found that after the wallet is connected, every time the dApp sends an `alph_requestNodeApi` request, it needs to be manually switched to the wallet so that the wallet can receive the request and process it. This problem did not occur when I tested it on an **_android physical_** machine. I don’t know if it’s a limitation of the iOS simulator.

3. The mobile wallet should return a response every time the request fails: https://github.com/alephium/alephium-frontend/blob/4ff8fd822ed998d6d2e8745c3ddeb2f176a298e2/apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx#L324, similar to the desktop wallet: https://github.com/alephium/alephium-frontend/blob/4ff8fd822ed998d6d2e8745c3ddeb2f176a298e2/apps/desktop-wallet/src/contexts/walletconnect.tsx#L304

@nop33 Please help confirm these issues when you are available, thank a lot!

### DIFF
--- a/packages/walletconnect/package.json
+++ b/packages/walletconnect/package.json
@@ -34,9 +34,9 @@
   "dependencies": {
     "@alephium/web3": "workspace:^",
     "@alephium/web3-wallet": "workspace:^",
-    "@walletconnect/sign-client": "2.5.2",
-    "@walletconnect/types": "2.5.2",
-    "@walletconnect/utils": "2.5.2",
+    "@walletconnect/sign-client": "2.10.6",
+    "@walletconnect/types": "2.10.6",
+    "@walletconnect/utils": "2.10.6",
     "eventemitter3": "^4.0.7"
   },
   "devDependencies": {
--- a/packages/walletconnect/src/constants.ts
+++ b/packages/walletconnect/src/constants.ts
@@ -34,3 +34,5 @@ export const RELAY_METHODS = [
 export const LOGGER = 'error'
 
 export const RELAY_URL = 'wss://relay.walletconnect.com'
+
+export const ALEPHIUM_DEEP_LINK = 'alephium://'
--- a/packages/walletconnect/src/provider.ts
+++ b/packages/walletconnect/src/provider.ts
@@ -138,18 +138,13 @@ export class WalletConnectProvider extends SignerProvider {
   }
 
   private async ping(topic: string) {
-    return await new Promise((resolve, reject) => {
-      const timeout = setTimeout(() => reject(new Error('Timeout')), 5 * 1000)
-      this.client
-        .ping({ topic })
-        .then((result) => {
-          clearTimeout(timeout)
-          resolve(result)
-        })
-        .catch((err) => {
-          clearTimeout(timeout)
-          reject(err)
-        })
+    const timeout = new Promise<never>((_, reject) => {
+      setTimeout(() => reject(new Error('Auto connect timeout')), 5 * 1000)
+    })
+    await Promise.race([this.client.ping({ topic }), timeout]).catch((err) => {
+      this.client.session.delete(topic, { code: 0, message: `Error: ${err}` })
+      this.session = undefined
+      this.account = undefined
     })
   }
 
--- a/packages/walletconnect/src/provider.ts
+++ b/packages/walletconnect/src/provider.ts
@@ -18,6 +18,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 import EventEmitter from 'eventemitter3'
 import { SessionTypes } from '@walletconnect/types'
 import SignClient from '@walletconnect/sign-client'
+import { isBrowser } from '@walletconnect/utils'
 import { getChainsFromNamespaces, getAccountsFromNamespaces, getSdkError } from '@walletconnect/utils'
 import {
   SignerProvider,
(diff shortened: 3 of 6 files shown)


polarker (approved):
LGTM

nop33:
Thanks for this work and your testing @Lbqds! Much appreciated!

> 1. The current desktop wallet automatically pops up when it receives a request, and the browser also displays a request to redirect to the desktop wallet.

Let me get this straight: You've been testing using a dApp on a desktop browser, on a computer that you have the desktop wallet installed, and also the android/ios simulator. When you connect the simulator to the desktop browser dApp via WalletConnect and request a signature, instead of only the simulator responding to this request, also the desktop wallet pops up. Is that the case?

First of all, let me make clear that we DO want the desktop wallet to automatically come to the foreground if it's connected via the "Desktop Wallet" connector to a desktop browser dApp. It's a feature, not a bug! 

At the moment, both [mobile wallet](https://github.com/alephium/alephium-frontend/blob/master/apps/mobile-wallet/app.config.js#L28) and [desktop wallet](https://github.com/alephium/alephium-frontend/blob/master/apps/desktop-wallet/public/electron.js#L26-L29) are listening to the same deep link protocol: `alephium://`. So, it would make sense that both apps are trying to respond **in the specific case where the mobile wallet has connected to a desktop browser dApp with the desktop wallet app installed on that desktop**.

A better solution (IMO) would be not to share the same protocol.

image

- The `@alephium/web3-react` lib redirects to `alephium://` when [using the "Desktop Wallet" connector button](https://github.com/alephium/alephium-web3/blob/master/packages/web3-react/src/utils/connector.ts#L114) 
- The `@alephium/walletconnect-provider` lib also redirects to `alephium://` when [using the "WalletConnect" connector button](https://github.com/alephium/alephium-web3/blob/master/packages/walletconnect/src/constants.ts#L38).

My opinion is that we should change the deep link of the [desktop wallet](https://github.com/alephium/alephium-frontend/blob/master/apps/desktop-wallet/public/electron.js#L26-L29) and the [connector](https://github.com/alephium/alephium-web3/blob/master/packages/web3-react/src/utils/connector.ts#L114) to `alephium-desktop://` and leave the deep link of the "WalletConnect" connector (and mobile wallet) to `alephium://`. Reason: The "WalletConnect" connector is more generic than the "Desktop Wallet" one. If more mobile wallets get developed by the community in the future, they could all use `alephium://` as their deep link and the Android system would let the user choose with which app they want to respond.

This would probably require the existing dApps to update their dependency on `@alephium/web3-react` if they want this problem fixed.

@Lbqds @mvaivre, WDYT?

> 2. When I used the ios simulator to test, I found that after the wallet is connected, every time the dApp sends an alph_requestNodeApi request, it needs to be manually switched to the wallet so that the wallet can receive the request and process it. This problem did not occur when I tested it on an android physical machine. I don’t know if it’s a limitation of the iOS simulator.

@mvaivre do you have some feedback on this? I remember you mentioned once that iOS has some limitations when it comes to background operations...

> 3. The mobile wallet should return a response every time the request fails

Good catch, thank you! I'll fix this today. https://github.com/alephium/alephium-frontend/issues/163

nop33:
> The current desktop wallet automatically pops up when it receives a request, and the browser also displays a request to redirect to the desktop wallet.

I understand this issue better now. And yes, I agree with @Lbqds that we should remove the line of code he recommends. Let me do some testing to confirm everything.

nop33:
As discussed on Slack, the preferred option is (1.b): Only redirect to the deep link if the dApp is running on a mobile browser. The reason is that changing the desktop wallet deep link scheme will break compatibility with the existing desktop wallet installations. The desktop wallet is considered more mature than the mobile.

1. If the dApp is running on a desktop browser, no matter if we are connected to it with the desktop or the mobile wallet, there's no need to redirect to the deep link:
   - The mobile wallet is a different device and reacts to WalletConnect messages
   - The desktop wallet can be brought to the foreground with the [existing implementation `app.show()`](https://github.com/alephium/alephium-frontend/blob/4ff8fd822ed998d6d2e8745c3ddeb2f176a298e2/apps/desktop-wallet/src/contexts/walletconnect.tsx#L186).
2. If the dApp is running on a mobile browser, chances are that we are connected with the mobile wallet (technically it's possible to also connect with the desktop wallet by copying the WC URL from the mobile browser and paste it in the desktop wallet, but we won't consider that cause it's a rare case). In this case, the SDK should redirect to the deep link so that the mobile wallet is brought to the foreground.

<details><summary>Click here for a more detailed summary of possible connections between our wallets and WC</summary>

I think the below list covers all possible connections of our wallets to all possible browser dApps:

The desktop wallet can connect to:
1. desktop browser dApps by:
    1. :star: using the “Desktop wallet” connector of web3-react
    1. using a WalletConnect URL (manual pasting) of the “WalletConnect” connector
1. (very unlikely case) mobile browser dApps by:
    1. using a WalletConnect URL (manual pasting) of the “WalletConnect” connector

The mobile wallet can connect to:
1. desktop browser dApps by:
    1. using the “WalletConnect” connector of web3-react by either
        1. :star: scanning the QR code
        1. manual pasting of the WalletConnect URL
1. mobile browser dApps by:
    1. using the “WalletConnect” connector of web3-react by either
        1. :star: clicking the “Connect” button (Android) to get redirected to the mobile wallet
        1. manual pasting of the WalletConnect URL

</details>

PR #283, opened 2023-12-09
Select session by network and group from alephium namespace

Lbqds:
In bridge, there may be multiple walletconnect connections of different chains at the same time, and we need to find the correct session.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@alephium/cli",
-  "version": "0.25.0",
+  "version": "0.25.1",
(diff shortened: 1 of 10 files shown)


polarker:
LGTM. Does this fix an existing bug?

polarker (approved):
LGTM. Does this fix an existing bug?

Lbqds:
> LGTM. Does this fix an existing bug?

Yes, I found this bug when testing the bridge. Bridge transfers require connecting two chains at the same time. If both chains are connected through walletconnect, we need to select the correct session when auto reconnecting.

PR #286, opened 2023-12-11
Update explorer-backend schemas to 1.16.0

tdroxler:
This adds all metadatas, for tokens/nfts

Required by FE team for the next update of the explorer

### DIFF
--- a/packages/web3/src/api/api-explorer.ts
+++ b/packages/web3/src/api/api-explorer.ts
@@ -136,6 +136,15 @@ export interface ExplorerInfo {
   lastFinalizedInputTime: number
 }
 
+export interface FungibleTokenMetadata {
+  /** @format 32-byte-hash */
+  id: string
+  symbol: string
+  name: string
+  /** @format uint256 */
+  decimals: string
+}
+
 export interface Hashrate {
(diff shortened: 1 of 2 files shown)


packages/web3/src/api/api-explorer.ts
```diff
@@ -136,6 +136,15 @@ export interface ExplorerInfo {
   lastFinalizedInputTime: number
 }
 
+export interface FungibleTokenMetadata {
+  /** @format 32-byte-hash */
+  id: string
+  symbol: string
+  name: string
+  /** @format uint256 */
+  decimals: string
```
nop33, line 145: 😢

tdroxler: yeah I know, but as we discussed it's safer for the backend to return a BigNumer as that's what we get from the node.

nop33: I know, I know, I'm still sad tho :P

nop33: I'll get over it, don't worry 😆

polarker (approved):
LGTM

PR #288, opened 2023-12-13
Add multisig key type

Lbqds:


packages/web3/src/utils/utils.ts
```diff
@@ -31,6 +31,8 @@ export const networkIds = ['mainnet', 'testnet', 'devnet'] as const
 export type NetworkId = (typeof networkIds)[number]
 
 const ec = new EC('secp256k1')
+const PublicKeyBytesLength = 33
+const MaxKeySize = 32
```
Lbqds, line 35: I initially tried to use `encodeI256` to encode numbers, but since this function is under the `contract` package, when referencing this function in `utils`, a strange error occurs due to circular reference, so I set the maximum number of keys to 32, as it is the max value that can be encoded with a single byte.

h0ngcha0: I wonder if we could use codec to do this now.

Lbqds: Yeah, I'll update this later 👍

h0ngcha0 (approved):
LGTM :+1:

PR #290, opened 2023-12-15
Clean api requests from history

Lbqds:


packages/walletconnect/src/provider.ts
```diff
@@ -322,6 +338,9 @@ export class WalletConnectProvider extends SignerProvider {
         chainId: this.permittedChain,
         topic: this.session?.topic
       })
+      if (!isSignRequest) {
+        this.cleanHistory()
```
Lbqds, line 342: We can't get the request id so we have to find it from history and delete it.

polarker (approved):
LGTM

PR #291, opened 2023-12-18
Fix clean history

Lbqds:


packages/walletconnect/src/provider.ts
```diff
@@ -218,9 +207,12 @@ export class WalletConnectProvider extends SignerProvider {
 
   // ---------- Private ----------------------------------------------- //
 
-  private cleanHistory() {
+  private cleanHistory(checkResponse: boolean) {
```
Lbqds, line 210: Sometimes there are multiple requests at the same time. We need to filter out those requests that have not received a response, otherwise walletconnect will throw an exception during processing.

packages/walletconnect/src/provider.ts
```diff
@@ -134,19 +134,8 @@ export class WalletConnectProvider extends SignerProvider {
       this.updateNamespace(this.session.namespaces)
     } else {
       this.updateNamespace(this.session.namespaces)
-      await this.ping(this.session.topic)
```
Lbqds, line 137: I want to remove ping for two reasons:
1. Before the ping request returns, the dApp cannot proceed to the next step, which is not very user-friendly, and the ETH walletconnect provider will not send the ping request either
2. Even if the wallet is not open when the dApp sends the request, the request will be saved to the walletconnect relayer server. Once the wallet is open and connected to the relay, the request will be received.

packages/walletconnect/src/provider.ts
```diff
@@ -134,19 +134,8 @@ export class WalletConnectProvider extends SignerProvider {
...
-  private async ping(topic: string) {
-    const timeout = new Promise<never>((_, reject) => {
-      setTimeout(() => reject(new Error('Auto connect timeout')), 5 * 1000)
-    })
-    await Promise.race([this.client.ping({ topic }), timeout]).catch((err) => {
-      this.session = undefined
-      this.account = undefined
-      return this.client.disconnect({ topic, reason: { code: 0, message: `Error: ${err}` } })
-    })
+    await this.client.core.relayer.messages.del(this.session.topic)
```
Lbqds, line 138: Every time a dApp sends a request, walletconnect stores the request message in the `messages` store. It will only be deleted if it is disconnected or the session expires.

PR #298, opened 2024-01-08
Support pool reward tx

Lbqds:


packages/web3/src/utils/exchange.ts
```diff
@@ -51,17 +55,29 @@ export function isSimpleTransferTokenTx(tx: Transaction): boolean {
...
-  let depositAmount = 0n
+export function getALPHDepositInfo(tx: Transaction): { targetAddress: Address; depositAmount: bigint }[] {
+  if (!isALPHTransferTx(tx)) {
+    return []
+  }
+  const inputAddresses: Address[] = []
+  for (const input of tx.unsigned.inputs) {
+    try {
+      const address = getAddressFromUnlockScript(input.unlockScript)
+      inputAddresses.push(address)
```
polarker, line 66: There might be duplicated addresses in tx inputs

packages/web3/src/utils/exchange.test.ts
```diff
@@ -24,7 +24,7 @@ import {
   getSenderAddress,
   getALPHDepositInfo,
   isSimpleALPHTransferTx,
-  isSimpleTransferTokenTx,
+  isSimpleTokenTransferTx,
```
polarker, line 27: How about removing these two `isSimple*` functions? They are not recommended for exchanges now.

test/exchange.test.ts
```diff
@@ -208,8 +208,8 @@ class Exchange {
   async handleBlock(block: node.BlockEntry, resolver: () => void) {
     for (const tx of block.transactions) {
       if (isSimpleALPHTransferTx(tx)) {
-        const { targetAddress, depositAmount } = getALPHDepositInfo(tx)
-        if (this.hotAddresses.includes(targetAddress)) {
+        const infos = getALPHDepositInfo(tx).filter((v) => this.hotAddresses.includes(v.targetAddress))
```
polarker, line 211: We should replace `isSimpleALPHTransferTx` with the new validation function.

polarker (commented):
Let's transfer to every address 1 ALPH from the testWallet in the integration test, so that we test the pool reward case there.

PR #300, opened 2024-01-12
Support debugging mode

### LINKED ISSUE #296: Support debugging mode
When debug mode is configure in `alephium.config.ts`, all of the HTTP requests would be logged, like Wallet connect

Lbqds:
Resolve #296

### DIFF
--- a/packages/web3/src/api/types.ts
+++ b/packages/web3/src/api/types.ts
@@ -17,6 +17,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
 import { ZERO_ADDRESS } from '../constants'
+import { isDebugModeEnabled } from '../debug'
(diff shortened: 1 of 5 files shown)


packages/web3/src/api/types.ts
```diff
@@ -267,20 +268,51 @@ export interface ApiRequestArguments {
...
       }
     }
   }
 }
 
 export async function request(provider: Record<string, any>, args: ApiRequestArguments): Promise<any> {
   const call = provider[`${args.path}`][`${args.method}`] as (...any) => Promise<any>
-  return call(...args.params)
+  logRequest(args.path, args.method, args.params)
+  const response = await call(...args.params)
```
polarker, line 313: Let's log the exception as well

Issue #301, opened 2024-01-14
Improve Dev Experience


polarker:
Feel free to share any specific suggestions to enhance the overall experience, especially with examples. The dev team is currently busy, but we'll soon have more time to fine-tune things. Your input is valuable!

polarker:
Error message lacks clarity when function parameters are not provided: https://discord.com/channels/747741246667227157/948144672402972682/1196050728909807716

ross-weir:
**Suggest fixes for errors**

Not sure how tricky to implement but maybe some low hanging fruit, i.e incorrectly naming builtin funcs, etc:

```
assert(1 == 1, 5)
```

Current error:
```
Failed to compile, error: Error: [API Error] - org.alephium.ralph.Compiler$Error: Function assert does not exist
```

With suggestion:
```
Failed to compile, error: Error: [API Error] - org.alephium.ralph.Compiler$Error: Function assert does not exist, did you mean to use builtin function "assert!"?
```

**Clearer error when declaring contract attributes in wrong order**

For example declaring `enum` before `event` with this contract:
```
Contract Example() {
  enum Error {
    Assert = 0
  }

  event StateChanged(old: U256, new: U256)

  pub fn helloWorld() -> () {
    assert!(1 == 1, Error.Assert)
  }
}
```

Throws this error that doesn't indicate the issue:
```
Failed to compile, error: Error: [API Error] - org.alephium.ralph.Compiler$Error: -- error (6:3): Syntax error
6 |  event StateChanged(old: U256, new: U256)
  |  ^^^^^^^^^^
  |  Expected "}"
  |-------------------------------------------------------------------------------------
  |Trace log: Expected multiContract:1:1 / rawContract:1:1 / "}":6:3, found "event Stat"
```

ross-weir:
**Implement user friendly error reporting**

Currently raw errors are shown including stacktraces, etc which likely isn't useful to contract devs most of the time, consider making error reporting more user friendly/concise.

Current error:

<details>
    <summary>Error output</summary>

```
Failed to compile, error: Error: [API Error] - org.alephium.ralph.Compiler$Error: -- error (6:3): Syntax error
6 |  event StateChanged(old: U256, new: U256)
  |  ^^^^^^^^^^
  |  Expected "}"
  |-------------------------------------------------------------------------------------
  |Trace log: Expected multiContract:1:1 / rawContract:1:1 / "}":6:3, found "event Stat"

[...]

PR #303, opened 2024-01-16
Remove cli stacktrace from cli error output

ross-weir:
Addresses comment: https://github.com/alephium/alephium-web3/issues/301#issuecomment-1891104267

New output from `compile` command:

```
Full node version: v2.5.7
Compiling contracts in folder "contracts"
✘ Failed to compile, error: [API Error] - org.alephium.ralph.Compiler$Error: -- error (9:3): Syntax error
9 |  event Hello(
  |  ^^^^^^^^^^
  |  Expected "}"
  |-------------------------------------------------------------------------------------
  |Trace log: Expected multiContract:1:1 / rawContract:1:1 / "}":9:3, found "event Fund"
```

### DIFF
--- a/packages/cli/cli_internal.ts
+++ b/packages/cli/cli_internal.ts
@@ -16,7 +16,15 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { Project, web3, NetworkId, networkIds, validateNFTBaseUri, enableDebugMode } from '@alephium/web3'
+import {
+  Project,
+  web3,
+  NetworkId,
+  networkIds,
+  validateNFTBaseUri,
+  enableDebugMode,
+  isDebugModeEnabled
+} from '@alephium/web3'
 import { program } from 'commander'
 import { run as runJestTests } from 'jest'
 import path from 'path'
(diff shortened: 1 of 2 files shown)


packages/cli/cli_internal.ts
```diff
@@ -209,6 +216,7 @@ nftCommand
   .option('-d, --localDir <directory-of-local-images>', 'directory of local images to be uploaded')
   .option('-i, --ipfsDir <ipfs-directory-of-uploaded-images>', 'IPFS directory to upload the images')
   .option('-m, --metadataFile <metadata-file>', 'file to store the metadata of the uploaded images')
+  .option('--debug', 'show detailed debug information such as error stack traces')
```
polarker, line 219: Let's call `enableDebugMode` as well when this is enabled, so that all of the API queries would be logged as well.

Not sure the log would look nice, otherwise please feel free to improve it.

ross-weir: Done

packages/cli/cli_internal.ts
```diff
@@ -168,7 +174,7 @@ program
       const toIndex = tryGetScriptIndex(options.to)
       await deployAndSaveProgress(config, networkId, fromIndex, toIndex)
     } catch (error) {
-      program.error(`Failed to deploy contracts, error: ${(error as Error).stack}`)
+      program.error(`✘ Failed to deploy contracts, error: ${buildErrorOutput(error, options.debug)}`)
```
polarker, line 177: If debug is enabled via the config file, should we also show the debug info? It looks good to keep the two debug flag consistent.

ross-weir: > If debug is enabled via the config file, should we also show the debug info? It looks good to keep the two debug flag consistent.

Yeah good point, updated

PR #308, opened 2024-02-10
Show Debug Info

h0ngcha0:


packages/web3/src/contract/contract.ts
```diff
@@ -1055,9 +1064,9 @@ export class Contract extends Artifact {
     return signerParams
   }
 
-  buildByteCodeToDeploy(initialFields: Fields): string {
+  buildByteCodeToDeploy(initialFields: Fields, isDevnet: boolean): string {
     try {
-      return ralph.buildContractByteCode(this.bytecode, initialFields, this.fieldsSig)
+      return ralph.buildContractByteCode(isDevnet ? this.bytecodeDebug : this.bytecode, initialFields, this.fieldsSig)
```
h0ngcha0, line 1069: Deploy `bytecodeDebug` to devnet, so we can print the debug message.

docker/docker-compose.yml
```diff
@@ -31,7 +31,7 @@ services:
         condition: service_healthy
 
   alephium:
-    image: alephium/dev-alephium:2.8.0
+    image: liuhongchao/dev-alephium:test-debug
```
h0ngcha0, line 34: Will update when we have the new full node image.

packages/cli/src/deployment.ts
```diff
@@ -317,8 +317,9 @@ function createDeployer<Settings = unknown>(
     params: DeployContractParams<P>,
     taskTag?: string
   ): Promise<DeployContractResult<T>> => {
+    const isDevnet = network.networkId !== 0 && network.networkId !== 1
```
polarker, line 320: Maybe we use `networkId % 3 == 2` following the Scala code. `%3 = 0, 1, 2` would be `mainnet, testnset, and devnet`.

polarker: As discussed, let's keep this for now until the next network upgrade

docker/devnet.conf
```diff
@@ -26,7 +26,7 @@ alephium.consensus.block-target-time = 10 millis
 alephium.consensus.uncle-dependency-gap-time = 0 seconds
 alephium.network.leman-hard-fork-timestamp = 1643500800000 // GMT: 30 January 2022 00:00:00
 
-alephium.network.network-id = 5
+alephium.network.network-id = 4
```
h0ngcha0, line 29: Will address the config issue in a seperate PR
This requires an update to full node devnet config template, which affected a lot of tests.

polarker (approved):
LGTM

PR #309, opened 2024-02-13
Fix devnet url

polarker:


packages/cli/templates/base/alephium.config.ts
```diff
@@ -27,7 +27,7 @@ const defaultSettings: Settings = {
 const configuration: Configuration<Settings> = {
   networks: {
     devnet: {
-      nodeUrl: 'http://localhost:22973',
+      nodeUrl: 'http://127.0.0.1:22973',
```
nop33, line 30: Why was this change needed? Do you think that the wallets should also use `127.0.0.1` instead of `localhost` in the network presets?

PR #313, opened 2024-02-20
Fix NFT metadata attributes type

nop33:
I believe the `attributes` type is wrong. They way it's currently defined, it can only be an array of 1 element whos type is 
```ts
{
  trait_type: string
  value: string | number | boolean
}
```

Instead, what we want, is an array where EACH element is of that type.


h0ngcha0 (approved):
LGTM, thanks!

PR #315, opened 2024-02-21
Reposition and add file name to compilation error

tdroxler:
This PR takes the error message coming from the node compilation and shift the index, as well as adding the file name to the error.
Received from node:
image
after:
image

### DIFF
--- a/packages/web3/src/contract/contract.ts
+++ b/packages/web3/src/contract/contract.ts
@@ -36,6 +36,7 @@ import {
   typeLength,
   getDefaultValue
 } from '../api'
+import { CompileProjectResult } from '../api/api-alephium'
 import {
   SignDeployContractTxParams,
   SignDeployContractTxResult,
@@ -63,6 +64,7 @@ import * as path from 'path'
 import { EventSubscribeOptions, EventSubscription, subscribeToEvents } from './events'
 import { ONE_ALPH } from '../constants'
 import * as blake from 'blakejs'
+import { parseError } from '../utils/error'
 
 const crypto = new WebCrypto()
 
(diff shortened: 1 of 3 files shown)


packages/web3/src/utils/error.ts
```diff
@@ -0,0 +1,77 @@
...
+    } else {
+      return newError
+    }
+  }
+}
+
+const errorRegex = /error \((\d+):(\d+)\):\s*(.*)\n\s*(\d+)\s*\|(.*)\n.*\|(.*)\n\s*\|(.*)(?:\n\s*\|(.*)\n\s*\|(.*))?/
+
+export function parseError(error: string): CompilationError | undefined {
+  const match = error.match(errorRegex)
```
h0ngcha0, line 50: [Nit] If for some reason this doesn't match the error regex, does it make sense to fall back to the original error? Otherwise I guess we might get regex error below.

packages/web3/src/utils/error.ts
```diff
@@ -0,0 +1,77 @@
...
+  ) {}
+
+  reformat(line: number, file: string): string {
+    const spaces = `${line}`.replace(/\d/g, ' ')
+    const newError = `${file} (${line}:${this.column}): ${this.errorType}
+${line} |${this.codeLine}
+${spaces} |${this.errorIndicator}
+${spaces} |${this.message}`
+
+    if (this.additionalLine1 && this.additionalLine2) {
```
h0ngcha0, line 39: Q: What is `additionalLine1` and `additionalLine2`? Is it footer?

packages/web3/src/contract/contract.ts
```diff
@@ -190,6 +192,31 @@ type CodeInfo = {
...
+    currentLine += source.sourceCode.split('\n').length
+    const endIndex = currentLine
+    return { sourceInfo: source, startIndex: startIndex, endIndex: endIndex }
+  })
+
+  const sourceInfo = sourceInfosWithLine.find((sourceInfoWithLine) => {
+    return line >= sourceInfoWithLine.startIndex && line <= sourceInfoWithLine.endIndex
+  })
+
+  if (sourceInfo === undefined) {
```
h0ngcha0, line 214: [Nit] `return sourceInfo`

h0ngcha0 (commented):
LGTM, left few minor comments

packages/web3/src/utils/error.ts
tdroxler: we throw it [here](https://github.com/alephium/alephium-web3/blob/1a9019ae2678ba60e9a81fdf49954966831e7892/packages/web3/src/contract/contract.ts#L477-L480) when we use the parsing function.

packages/web3/src/utils/error.ts
tdroxler: Some errors can have a "hint" footer, see [this example from the test](https://github.com/alephium/alephium-web3/blob/1a9019ae2678ba60e9a81fdf49954966831e7892/packages/web3/src/utils/error.test.ts#L39-L44) Those are "real" error generated by our node `CompilerSpec`

packages/web3/src/contract/contract.ts
tdroxler: ah indeed :sweat_smile: It's fixed

PR #323, opened 2024-02-28
Improve the return type of deployToDevnet

Lbqds:
In this PR, I changed the return type of `deployToDevnet` to:

````
{
  deployerAddress: string,
  contracts: Record<string, DeployContractExecutionResult>,
  scripts:Record<string, RunScriptResult>
}[]
````

But it is a bit different from `loadDeployments`. For `loadDeployments`, we generate the return type through the deployed contracts. For `deployToDevnet`, we don't know the return type until the user calls `deployToDevnet`, so the user needs to do some type conversion manually, here is an example: https://github.com/alephium/nextjs-template/commit/737cf8f10e5f119da63f96d769dd44907dc42a4f#diff-c44cb9b2c107e64a3654d247df87d052da7c9c4ee019f47ee0f1678fc7d933a7R27-R30

### CODE REVIEW

File: packages/cli/src/codegen.ts [hunk 1]
```diff
@@ -686,16 +700,61 @@ export function sortByName<T extends { artifact: { name: string } }>(artifacts:
   return artifacts.sort((a, b) => (a.artifact.name > b.artifact.name ? 1 : -1))
 }
 
+function genUtils(outDir: string) {
+  const utilsPath = path.join(outDir, 'utils.ts')
+  const source = `
+    import { ContractInstance, groupOfAddress } from '@alephium/web3'
+    import { DeployContractExecutionResult } from '@alephium/cli'
+
+    export function loadContractInstanceFromDeployments<T extends ContractInstance>(
```
Lbqds, line 709: Because there are dependencies in `@alephium/cli` that only work in nodejs env, placing this function within `@alephium/cli` would lead to frontend issues. So I've moved this function to a separately generated file.

PR #330, opened 2024-03-05
Upgrade WalletConnect deps

nop33:


polarker:
This is done in another PR now.

PR #336, opened 2024-03-19
Fix subscribe system events

Lbqds:


Lbqds:
`pnpm audit --fix` isn't working again.

PR #344, opened 2024-04-15
Import interfaces from std

Lbqds:
If the user already imported `std/nft_collection_interface`, there's no need to import `std/nft_interface` again


h0ngcha0 (approved):
LGTM :+1:

PR #345, opened 2024-04-18
Simple Transaction Method Call Without Explicit TxScript

h0ngcha0:
Support simple transaction method call without explicit `TxScript`, example:

- Primitive function parameters:
```
await add.transaction.destroy({ args: { caller: caller }, signer })
```

- Arrays and struct:
```
    await add.transaction.add2({
      args: {
        array1: [2n, 1n],
        address: signer.address,
        array2: [2n, 1n],
        addS: {
          a: 1n,
          b: [
            { a: 1n, b: [2n, 1n] },
            { a: 1n, b: [2n, 1n] }
          ]
        }
      },
      signer
    })
```

- asset transfer
```
add.transaction.createSubContractAndTransfer({
      args: {
        a: 1n,
        path: stringToHex('test-path'),
        subContractId: sub.contractInstance.contractId,
        payer: signerAddress
      },
      signer,
      attoAlphAmount: ONE_ALPH * 2n,
      tokens: [{ id: sub.contractInstance.contractId, amount: 200n }],
      approve: {
        attoAlphAmount: ONE_ALPH,
        tokens: [{ id: sub.contractInstance.contractId, amount: 200n }]
      }
    })
```

### DIFF
--- a/packages/web3/src/api/utils.test.ts
+++ b/packages/web3/src/api/utils.test.ts
@@ -16,6 +16,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
+import { enableDebugMode, groupOfAddress } from '..'
 import { ONE_ALPH } from '../constants'
 import { NodeProvider, isBalanceEqual, node } from './index'
 
--- a/packages/web3/src/contract/contract.ts
+++ b/packages/web3/src/contract/contract.ts
@@ -71,6 +71,7 @@ import { isContractDebugMessageEnabled } from '../debug'
 import {
   contract,
   compactUnsignedIntCodec,
+  compactSignedIntCodec,
   Method,
   LoadLocal,
   LoadImmFieldByIndex,
@@ -2279,7 +2280,7 @@ export async function signExecuteMethod<I extends ContractInstance, F extends Fi
 ): Promise<SignExecuteScriptTxResult> {
   const methodIndex = contract.contract.getMethodIndex(methodName)
   const functionSig = contract.contract.functions[methodIndex]
-  const bytecodeTemplate = encodeBytecodeTemplate(methodIndex, functionSig)
+  const bytecodeTemplate = encodeBytecodeTemplate(methodIndex, functionSig, contract.contract.structs)
 
   const fieldsSig = toFieldsSig(contract.contract.name, functionSig)
   const bytecode = ralph.buildScriptByteCode(
@@ -2304,28 +2305,35 @@ export async function signExecuteMethod<I extends ContractInstance, F extends Fi
   return await signer.signAndSubmitExecuteScriptTx(signerParams)
 }
 
-function encodeBytecodeTemplate(methodIndex: number, functionSig: FunctionSig): string {
+function encodeBytecodeTemplate(methodIndex: number, functionSig: FunctionSig, structs: Struct[]): string {
   const numberOfMethods = '01'
   const isPublic = '01'
   const modifier = '03'
   const argsLength = '00'
-  const localsLength = '00'
   const returnsLength = '00'
 
-  let functionParamsTemplates = ''
-  functionSig.paramNames.forEach((_, index) => {
-    functionParamsTemplates += '{' + (index + 1) + '}'
-  })
+  const [templateVarStoreLocalInstrs, templateVarsLength] = getTemplateVarStoreLocalInstrs(functionSig, structs)
 
-  // These two we need to calculate based on functionSig, it should be
-  // easy
-  const functionArgsNum = encodeU256Const(functionSig.paramNames.length)
-  const functionReturnNum = encodeU256Const(functionSig.returnTypes.length)
+  // -1 because the first template var is the contract
+  const functionArgsNum = encodeU256Const(templateVarsLength - 1)
+  const localsLength = (templateVarStoreLocalInstrs.length / 2).toString(16).padStart(2, '0')
 
-  // This is fixed, we need this as contract id
-  const contractTemplate = '{0}' // always the 1st argument
+  const templateVarLoadLocalInstrs = getTemplateVarLoadLocalInstrs(functionSig, structs)
+
+  const functionReturnTypesLength: number = functionSig.returnTypes.reduce(
+    (acc, returnType) => acc + ralph.typeLength(returnType, structs),
+    0
+  )
+  const functionReturnPopInstrs = '18'.repeat(functionReturnTypesLength)
+  const functionReturnNum = encodeU256Const(functionReturnTypesLength)
+
+  const contractTemplateVar = '{0}' // always the 1st argument
   const externalCallInstr = '01' + methodIndex.toString(16).padStart(2, '0')
-  const numberOfInstrs = (functionSig.paramNames.length + 4 + 0).toString(16).padStart(2, '0')
+  const numberOfInstrs = compactSignedIntCodec
+    .encodeI32(
+      templateVarStoreLocalInstrs.length + templateVarLoadLocalInstrs.length + functionReturnTypesLength + 4 // functionArgsNum, functionReturnNum, contractTemplate, externalCallInstr
+    )
+    .toString('hex')
 
   return (
     numberOfMethods +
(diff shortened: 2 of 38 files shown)


packages/web3/src/contract/ralph.ts
```diff
@@ -385,6 +385,28 @@ export function primitiveToByteVec(value: Val, type: string): Uint8Array {
...
+    const [baseType, size] = decodeArrayType(typ)
+    return size * typeLength(baseType, structs)
+  }
+
+  const struct = structs.find((s) => s.name === typ)
+  if (struct !== undefined) {
+    return struct.fieldTypes.reduce((acc, fieldType) => acc + typeLength(fieldType, structs), 0)
+  }
+
+  throw Error(`Invalid type ${typ}`)
```
Lbqds, line 403: The compiler will return the contract name instead of `ByteVec` for the contract type, so we should return 1 here.

packages/web3/src/contract/ralph.ts
```diff
@@ -385,6 +385,28 @@ export function primitiveToByteVec(value: Val, type: string): Uint8Array {
...
+}
+
+export function isPrimitiveType(typ: string): boolean {
+  return ['Bool', 'I256', 'U256', 'ByteVec', 'Address'].includes(typ)
```
Lbqds, line 407: We have already defined the primitive types here: https://github.com/alephium/alephium-web3/blob/1325c8cb7a071820bcb2ea3d92539d2d2bebaa32/packages/web3/src/api/types.ts#L30
maybe we can reuse this one.

packages/web3/src/contract/ralph.ts
```diff
@@ -466,7 +488,7 @@ export function buildScriptByteCode(
 ): string {
   const allFields = flattenFields(fields, fieldsSig.names, fieldsSig.types, fieldsSig.isMutable, structs)
   return bytecodeTemplate.replace(scriptFieldRegex, (_, fieldIndex: string) => {
-    const field = allFields[`${fieldIndex}`]
+    const field = allFields[parseInt(fieldIndex)]
```
Lbqds, line 491: Question: why do we need to change this?

packages/web3/src/contract/contract.ts
```diff
@@ -1758,6 +1773,19 @@ export interface CallContractResult<R> {
...
 
+export interface SignExecuteContractMethodParams<T extends Arguments = Arguments> {
+  args: T
+  signer: SignerProvider
+  attoAlphAmount?: Number256
+  tokens?: Token[]
+  approve?: {
+    attoAlphAmount?: Number256
+    tokens?: Token[]
+  }
```
Lbqds, line 1784: Question: is it possible to only use one `attoAlphAmount` and `tokens` arg?

h0ngcha0: Great question. I am hesitating on this as well. 
For `attoAlphAmount`, we often include the extra amount for gas, so we can't automatically approve everything. 
For `tokens`, maybe we can automatically approve everything, but there are also scenarios where users just want to provide tokens in the transaction inputs without spending it (e.g. an NFT holder should be able to do something in a smart contract without transferring it)

But i do feel it's a bit duplicated to have both `approve` and `attoAlphAmount` & `tokens`, any suggestions for better interface is much appreciated.

packages/web3/src/contract/contract.ts
```diff
@@ -2260,6 +2288,237 @@ export async function callMethod<I extends ContractInstance, F extends Fields, A
...
+  return approveAlphInstrs
+}
+
+function getApproveTokensInstrs(tokens?: Token[]): string[] {
+  const approveTokensInstrs: string[] = []
+  if (tokens) {
+    tokens.forEach((token) => {
+      const tokenId = byteStringCodec.encodeBuffer(Buffer.from(token.id, 'hex'))
+      const tokenAmount = encodeU256Const(BigInt(token.amount))
+      approveTokensInstrs.push('1440' + tokenId.toString('hex'))
```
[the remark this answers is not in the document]
h0ngcha0: Great catch :+1:

packages/web3/src/contract/ralph.ts
h0ngcha0: reverted now

packages/web3/src/contract/contract.ts
Lbqds: It seems like users don't need to consider the gas fee. The full node will select utxos for `attoAlphAmount + gasFee`: https://github.com/alephium/alephium/blob/master/flow/src/main/scala/org/alephium/flow/core/UtxoSelectionAlgo.scala#L146
but as you said, sometimes users need to approve extra dust amount, this means the script will approve more ALPH than before.

h0ngcha0: I suppose one approach is to default the "approve" value the `attoAlphAmount` and `tokens` amount, so in most cases we don't need to specify `approve`. But if we want to be explicit we have the option to do that. WDTY?

Lbqds: That sounds great 👍

polarker: Let's keep things straightforward and stick with attoAlphAmount and tokens for now. Approving a bit more dustAmount shouldn't cause any harm. When building the transaction, attoAlphAmount and tokens indicate how much of the assets the user wants to use, not how much is ultimately used in the transaction.

I recommend removing `approve` entirely for now and only adding it later if necessary.

@h0ngcha0, can you confirm that gas fee is not included in attoAlphAmount? It should not be, as including it would make things too complicated.

h0ngcha0: Yeah, gas fee is not included in `attoAlphAmount`.
Will remove `approve` now :+1:

packages/web3/src/api/utils.test.ts
```diff
@@ -16,6 +16,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
+import { enableDebugMode, groupOfAddress } from '..'
```
Lbqds, line 19: [Nit]: It seems like this is not used.

Lbqds (approved):
LGTM 👍

packages/web3/src/contract/contract.ts
```diff
@@ -2409,7 +2409,7 @@ function getApproveTokensInstrs(tokens?: Token[]): string[] {
     tokens.forEach((token) => {
       const tokenId = byteStringCodec.encodeBuffer(Buffer.from(token.id, 'hex'))
       const tokenAmount = encodeU256Const(BigInt(token.amount))
-      approveTokensInstrs.push('1440' + tokenId.toString('hex'))
+      approveTokensInstrs.push('14' + tokenId.toString('hex'))
```
polarker: Magic number

packages/web3/src/contract/ralph.ts
```diff
@@ -503,7 +521,7 @@ function _encodeField<T>(fieldName: string, encodeFunc: () => T): T {
     return encodeFunc()
   } catch (error) {
     if (error instanceof Error) {
-      throw new Error(`Invalid ${fieldName}, error: ${error.message}`)
+      throw new Error(`Invalid ${fieldName}, error: ${error.message} `)
```
polarker, line 524: Why do we add an empty space?

packages/web3/src/contract/ralph.ts
```diff
@@ -555,7 +573,7 @@ export function encodeContractField(tpe: string, value: Val): Uint8Array {
     case 'Address':
       return encodeVmAddress(toApiAddress(value))
     default:
-      throw Error(`Expected primitive type, got ${tpe}`)
+      throw Error(`Expected primitive type, got ${tpe} `)
```
polarker, line 576: Why do we add an empty space?

polarker:
@h0ngcha0 Let's resolve the conflicts.

Please also check if we use u32/u256 encoding when i32/i256 is needed

polarker:
To keep the naming consistent, maybe we should use `call` and `txCall`

packages/web3/src/contract/contract.ts
h0ngcha0: Fixed, thanks :+1:

packages/web3/src/contract/ralph.ts
h0ngcha0: Caused by the editor config, fixed, thanks.

polarker:
@h0ngcha0 There are still a few places to fix if you search for `fromU32`. Another way to check is to comment out all of the `U32` related code and see if the package would compile and the tests could pass.

polarker:
@h0ngcha0 Let's maybe remove `U32` related code so as to avoid such issues completely.

h0ngcha0:
> We don't use U32 in the code as Java does not support signed integers.

Thanks. that makes great sense. I removed `compactUnsignedIntCodec.fromU32` and all the places where we use `u32` in other codecs now.

polarker:
@h0ngcha0 Let's remove `toU32` and `decodeU32` as well

h0ngcha0:
> @h0ngcha0 Let's remove `toU32` and `decodeU32` as well

Both are removed now.

PR #350, opened 2024-05-10
Gen code for maps

Lbqds:
To test Ralph map, users need to define `contains` and `get` methods in the contract, which is not very convenient. This PR generates two methods(`contains` and `get`) for each map variable.


h0ngcha0 (approved):
Awesome :+1:

polarker (approved):
LGTM 👍

PR #351, opened 2024-05-13
Update sdk to support rhone upgrade

Lbqds:
1. Remove unnecessary fields from the artifact
2. Add the `@using(methodSelector = false)` annotation to existing std interfaces
3. Add a `skipSaveArtifacts` flag in the CLI:
    1. If `skipSaveArtifacts` is enabled or the contracts have been deployed on the mainnet, we won't regenerate bytecode for those unchanged contract
    2. If the existing contract code is changed, we will regenerate bytecode for the contract
    3. If the existing contract code remains unchanged but new contracts are added, we will only regenerate bytecode for the new contracts

### DIFF
--- a/packages/web3/src/contract/contract.ts
+++ b/packages/web3/src/contract/contract.ts
@@ -2069,24 +2069,36 @@ export async function testMethod<
 }
 
 export class RalphMap<K extends Val, V extends Val> {
+  private readonly groupIndex: number
   constructor(
     private readonly parentContract: Contract,
-    private readonly parentInstance: ContractInstance,
+    private readonly parentContractId: HexString,
     private readonly mapName: string
-  ) {}
+  ) {
+    this.groupIndex = groupOfAddress(addressFromContractId(parentContractId))
+  }
 
   async get(key: K): Promise<V | undefined> {
-    return getMapItem(this.parentContract, this.parentInstance, this.mapName, key)
+    return getMapItem(this.parentContract, this.parentContractId, this.groupIndex, this.mapName, key)
   }
 
   async contains(key: K): Promise<boolean> {
     return this.get(key).then((v) => v !== undefined)
   }
+
+  toJSON() {
+    return {
+      parentContractId: this.parentContractId,
+      mapName: this.mapName,
+      groupIndex: this.groupIndex
+    }
+  }
 }
 
 export async function getMapItem<R extends Val>(
   parentContract: Contract,
-  parentInstance: ContractInstance,
+  parentContractId: HexString,
+  groupIndex: number,
   mapName: string,
   key: Val
 ): Promise<R | undefined> {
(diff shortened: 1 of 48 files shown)


packages/web3/src/contract/contract.ts
```diff
@@ -821,19 +862,22 @@ export class Project {
...
     const fullNodeVersion = defaultFullNodeVersion ?? (await provider.infos.getInfosVersion()).version
     const sourceFiles = await Project.loadSourceFiles(projectRootDir, contractsRootDir)
     const { errorOnWarnings, ...nodeCompilerOptions } = { ...DEFAULT_COMPILER_OPTIONS, ...compilerOptionsPartial }
     const projectArtifact = await ProjectArtifact.from(projectRootDir)
+    const changedSources = projectArtifact?.getChangedSources(sourceFiles) ?? sourceFiles
     if (
       projectArtifact === undefined ||
-      projectArtifact.needToReCompile(nodeCompilerOptions, sourceFiles, fullNodeVersion)
+      projectArtifact.needToReCompile(nodeCompilerOptions, fullNodeVersion) ||
+      changedSources.length > 0
```
h0ngcha0, line 877: Q: Shall we check `skipSaveArtifacts` here? I suppose when `skipSaveArtifacts` is false and project is deployed on mainnet, we still want to potentially regenerate the code.

packages/web3/src/contract/contract.ts
```diff
@@ -531,9 +543,32 @@ export class Project {
...
+    for (const [_, script] of this.scripts) {
+      await saveToFile(script)
+    }
+    await this.saveStructsToFile()
+    await this.saveProjectArtifact(projectRootDir, skipSaveArtifacts, changedSources)
+  }
+
+  private async saveProjectArtifact(projectRootDir: string, skipSaveArtifacts: boolean, changedSources: SourceInfo[]) {
+    if (skipSaveArtifacts) {
+      // we should not update the `codeHashDebug` if the `skipSaveArtifacts` is enabled
```
h0ngcha0, line 560: Q: Shall we update `codeHashDebug` when the source code is updated even if `skipSaveArtifacts` is enabled?

Lbqds: Yes, this is how it works now.

If the `skipSaveArtifacts` is true, then we will check if the contract code changed by the `changedSource`, if the contract code is changed, then we will use the new `codeHashDebug`.

packages/web3/src/contract/contract.ts
Lbqds: If the project is deployed on mainnet, then the `skipSaveArtifacts` should be true, we have already checked this in the CLI.

If the `skipSaveArtifacts` is true and the contract code is unchanged, then we won't regenerate bytecode, but If the contract code is changed, we will regenerate bytecode.

h0ngcha0: I see, thanks!

h0ngcha0 (approved):
LGTM

polarker (approved):
LGTM

PR #356, opened 2024-05-24
Add TxScript Codec

h0ngcha0:
As requested by the Mobula team


packages/web3/src/codec/txscript-codec.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export class TxScriptCodec implements Codec<DecodedTxScript> {
+  parser = Parser.start().nest('methods', {
+    type: methodsCodec.parser
+  })
+
+  encode(input: DecodedTxScript): Buffer {
+    return methodsCodec.encode(input.methods.value)
+  }
```
Lbqds, line 40: Does it make sense to add an `encodeTxScript(script: TxScript)` method?

Lbqds (approved):
LGTM 👍

packages/web3/src/codec/txscript-codec.ts
h0ngcha0: Good idea, added now.

packages/web3/src/codec/txscript-codec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export interface DecodedTxScript {
+  methods: DecodedArray<DecodedMethod>
+}
+
+export interface TxScript {
+  methods: Method[]
+}
+
+export class TxScriptCodec implements Codec<DecodedTxScript> {
```
polarker, line 34: Is this used in transaction codec?

h0ngcha0: Great catch, we already have a Script codec, i will merge these two.

PR #357, opened 2024-05-27
Don't report warnings if the contract is not changed

Lbqds:
If the `skipSaveArtifacts` is true and the contract is not changed, we should ignore the compiler warnings.


h0ngcha0 (approved):
LGTM :+1:

polarker (approved):
LGTM 👍

PR #358, opened 2024-05-29
Move `Project.build` to the cli

Lbqds:
Unit tests don't need to call `Project.build` anymore.


packages/cli/src/project.ts
```diff
@@ -0,0 +1,797 @@
...
+    if (sourceInfos.length === 0 || contractAndScriptSize === 0) {
+      throw new Error('Project have no source files')
+    }
+    return sourceInfos.sort((a, b) => a.type - b.type)
+  }
+
+  static readonly DEFAULT_CONTRACTS_DIR = 'contracts'
+  static readonly DEFAULT_ARTIFACTS_DIR = 'artifacts'
+
+  static async build(
```
polarker, line 748: Let's rename this as `compile` so that devs don't use it the way as before

packages/cli/src/project.ts
```diff
@@ -0,0 +1,797 @@
...
+        contractsRootDir,
+        artifactsRootDir,
+        errorOnWarnings,
+        nodeCompilerOptions,
+        changedSources,
+        skipSaveArtifacts
+      )
+    }
+    // we need to reload those contracts that did not regenerate bytecode
+    Project.currentProject = await Project.loadArtifacts(
```
polarker, line 785: Do we still need `Project.currentProject`?

Lbqds: No, we don't need it anymore, removed

polarker (approved):
LGTM 👍

PR #359, opened 2024-05-30
Add the force recompile flag

Lbqds:


packages/cli/src/types.ts
```diff
@@ -33,14 +33,13 @@ import {
...
 import { getConfigFile, loadConfig } from './utils'
-import path from 'path'
 import { Project } from './project'
 
 export interface Network<Settings = unknown> {
   networkId?: number
   nodeUrl: string
   privateKeys: string[] | string
-  deploymentStatusFile?: string
+  deploymentFile?: string
```
Lbqds, line 42: We already support user-defined deployment file paths, so we don't need to introduce a new deployment config.

PR #361, opened 2024-06-04
Support custom connect provider

Lbqds:
Here is an example: https://github.com/alephium/nextjs-template/tree/custom-connect-provider


polarker (approved):
LGTM 👍

PR #363, opened 2024-06-04
Replace Buffer with Uint8Array in codec

Lbqds:
Currently, our codec uses `Buffer`, which is imported from `buffer/`. But it is not compatible with `Uint8Array`.


packages/web3/src/address/address.ts
```diff
@@ -20,9 +20,9 @@ import { ec as EC } from 'elliptic'
...
 import blake from 'blakejs'
-import bs58 from './bs58'
-import djb2 from './djb2'
-import { binToHex, hexToBinUnsafe } from './utils'
+import bs58 from '../utils/bs58'
+import djb2 from '../utils/djb2'
+import { binToHex, hexToBinUnsafe } from '../utils'
 import { KeyType } from '../signer'
 import { MultiSig, lockupScriptCodec } from '../codec/lockup-script-codec'
 import { Buffer } from 'buffer/'
```
nop33, line 28: Is this still needed?

packages/web3/src/codec/script-codec.test.ts
```diff
@@ -16,6 +16,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
+import { Buffer } from 'buffer/'
```
nop33, line 19: and this?

packages/web3/src/codec/bytestring-codec.test.ts
```diff
@@ -21,15 +21,15 @@ import { compactUnsignedIntCodec } from './compact-int-codec'
 
 describe('Encode & decode bytestring', function () {
   it('should encode & decode bytestring', function () {
-    success(Buffer.from(''))
-    success(Buffer.from('Alephium is great!'))
+    success(new Uint8Array([]))
+    success(new Uint8Array(Buffer.from('Alephium is great!')))
```
nop33, line 25: `Buffer` Could probably be removed from here as well?

nop33 (commented):
Nice job man!

Just my 2 cents, once the need for `Buffer` is removed, how about cleaning up the webpack polyfills as well?

- https://github.com/alephium/alephium-web3/blob/replace-buffer-with-uint8array-in-codec/packages/web3/webpack.config.js#L28-L30
- https://github.com/alephium/alephium-web3/blob/replace-buffer-with-uint8array-in-codec/packages/web3/webpack.config.js#L44

polarker:
@Lbqds Yes. Let's take the chance to remove `Buffer` completely from the project, not just `codec`.

Lbqds:
Replaced all `Buffer` with `Uint8Array` in the web3 package.

polarker (approved):
LGTM!

PR #369, opened 2024-06-11
Check if the contract method indices are the same

Lbqds:


packages/cli/src/project.ts
```diff
@@ -436,6 +436,22 @@ export class Project {
...
+    let oldArtifact: Contract
+    try {
+      oldArtifact = await Contract.fromArtifactFile(artifactPath, '', '')
+    } catch (error) {
+      throw new Error(`Failed to load contract artifact, error: ${error}, contract: ${newArtifact.sourceInfo.name}`)
+    }
+    newArtifact.artifact.functions.forEach((newFuncSig, index) => {
+      const oldFuncSig = oldArtifact.functions[`${index}`]
+      if (oldFuncSig.name !== newFuncSig.name) {
+        throw new Error(`Function ${newFuncSig.name} has invalid index ${index}`)
```
polarker, line 450: `The newly compiled contract ${newfuncSig.name} has different method indexes compared to the existing compilation.`

polarker (approved):
LGTM 👍 One minor comment

packages/cli/src/project.ts
```diff
@@ -447,7 +447,9 @@ export class Project {
     newArtifact.artifact.functions.forEach((newFuncSig, index) => {
       const oldFuncSig = oldArtifact.functions[`${index}`]
       if (oldFuncSig.name !== newFuncSig.name) {
-        throw new Error(`Function ${newFuncSig.name} has invalid index ${index}`)
+        throw new Error(
+          `The newly compiled contract ${newArtifact.artifact.name} has different method indexes compared to the existing compilation.`
```
polarker, line 451: more: `the existing deployment on mainnet/testnet`

PR #372, opened 2024-06-14
Remove references to Project from token.test.ts

DevRozaDev:
Citing @_muchen_:

>We moved the Project from @alephium/web3 to @alephium/cli in the latest version, and it is not exported from @alephium/cli. You don't need to call Project.build in testing anymore.


Lbqds (approved):
Thanks a lot!

Issue #362, opened 2024-06-04
Using @alephium/web3-wallet in react-native fails


gilles-hemmerle:
## Context

I'm currently trying to implement the  bridge in sezame wallet (react native application).
I imported wormhole-sdk and I try to use the bridge directly through the wallet.

I need to use `transferLocalTokenFromAlph` which expects a SignerProvider as a first params. So I figured out in your documentation that we can create a SignerProvider from the `PrivateKeyWallet` class which is exposed in the `@alephium/web3-wallet project`.

## Environnement 

- node 18.19.0
- yarn 1.22.21

**What I did**

I imported the following dependencies :

```json
    "@alephium/web3": "^0.42.0",
    "@alephium/web3-wallet": "^0.42.0",
    "@alephium/wormhole-sdk": "^0.2.5",
```

However, when I want to import the PrivateKeyWallet like this :

`import {PrivateKeyWallet} from '@alephium/web3-wallet';`

 I get the following error when running the app in a ios device

`npx react-native start --reset-cache` and then executing the app : `yarn ios`

error: Error: While trying to resolve module `@alephium/web3-wallet` from file `sezame/wallet-react-native/src/screens/bridge/bridge-screen.tsx`, the package `sezame/wallet-react-native/node_modules/@alephium/web3-wallet/package.json` was successfully found. However, this package itself specifies a `main` module field that could not be resolved (`sezame/wallet-react-native/node_modules/@alephium/web3-wallet/dist/alephium-web3.min.js`. Indeed, none of these files exist:

![image](https://github.com/alephium/alephium-web3/assets/2291800/b783862f-ba0b-4d8c-8fa0-2739fe0ac173)

## Repo to desmonstrate the issue

https://github.com/gilles-hemmerle/alephium-web3-wallet-issue

## Questions

* Anyone does have the same issue?
* Did you already try to implement the bridge in your rect native wallets, if yes, do you have a git showing this?
* I'm using the womhole sdk `transferLocalTokenFromAlph` function. The first argument requires a `SignerProvider`. But if I pass a `PrivateKeyWallet`, it complains : 'PrivateKeyWallet' is not assignable to the 'SignerProvider' type parameter. Howver,  in the source files I see `class PrivateKeyWallet extends SignerProviderSimple` and `abstract class SignerProviderSimple extends SignerProvide`, so `transferLocalTokenFromAlph` should accept a `PrivateKeyWallet`, do I have some inompatibility between the dependencies i'm using?

[...]

PR #377, opened 2024-06-20
Fix web3-wallet entry points

nop33:
So there are 2 problems with `@alephium/web3-wallet`.

The first one is that the [name of the Webpack minified output](https://github.com/alephium/alephium-web3/blob/v0.45.0/packages/web3-wallet/webpack.config.js#L44) is `alephium-web3-wallet.min.js`, but in the `package.json` it is [referred wrongly](https://github.com/alephium/alephium-web3/blob/v0.45.0/packages/web3-wallet/package.json#L11) as `alephium-web3.min.js`.

The second is that Webpack was not being executed so the minified file that RN expected was never there.

This PR should be addressing both issues.

Potentially closes https://github.com/alephium/alephium-web3/issues/362

### DIFF
--- a/packages/web3-wallet/package.json
+++ b/packages/web3-wallet/package.json
@@ -30,7 +30,7 @@
   },
   "author": "Alephium dev <dev@alephium.org>",
   "scripts": {
-    "build": "rm -rf dist/* && npx tsc --build .",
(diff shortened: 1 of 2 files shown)


Lbqds (approved):
LGTM 👍

PR #389, opened 2024-07-04
Fix minor bugs

Lbqds:


packages/web3/src/codec/lockup-script-codec.ts
```diff
@@ -17,15 +17,15 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+import { Codec, fixedSizeBytes } from './codec'
 import { ArrayCodec, DecodedArray } from './array-codec'
 
 export interface PublicKeyHash {
   publicKeyHash: Uint8Array
 }
 
 class PublicKeyHashCodec implements Codec<PublicKeyHash> {
-  parser = Parser.start().buffer('publicKeyHash', { length: 32 })
+  parser = fixedSizeBytes('publicKeyHash', 32)
```
Lbqds, line 28: We are encountering an issue with the `binary-parser` library, which causes the `lockupScriptCodec.decode(hexToBinUnsafe('01ca40da2a3f'))` function to return a run out of memory error:

1. Because the prefix is `1`, it attempts to decode `P2MPKH`.
2. The decoded array size is huge. And even if `from` and `to` are out of the index, `Uint8Array.subarray(from, to)` does not throw an exception but returns an empty `Uint8Array`. Moreover, `binary-parser` does not check the length during runtime, leading to the allocation of a very large array and resulting in the `out of memory` error.

I feel like `binary-parser` lacks flexibility. It does not provide a direct way to check the size of the data. Currently it can only be checked through the `wrapper`.

polarker (approved):
LGTM 👍

PR #395, opened 2024-07-11
Add a silent flag to remove deployment log output

Lbqds:
By default, the `CLI` will output the deployment log. Users can use the `--silent` flag to remove the deployment log output, which is similar to Hardhat.

### DIFF
--- a/packages/cli/src/deployment.ts
+++ b/packages/cli/src/deployment.ts
@@ -323,14 +323,25 @@ function getTaskId(code: Contract | Script, taskTag?: string): string {
   return `${code.name}:${taskTag}`
 }
 
-function debugLog(enableDebugMode: boolean, message: string) {
-  if (enableDebugMode) {
-    console.log(message)
+function getDebugLogger(enableDebugMode: boolean): (string) => void {
+  return (message: string): void => {
+    if (enableDebugMode) {
+      console.log(message)
+    }
+  }
+}
+
+function getDeploymentLogger(silent: boolean): (string) => void {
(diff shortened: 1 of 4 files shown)


packages/cli/src/deployment.ts
```diff
@@ -369,16 +387,16 @@ function createDeployer<Settings = unknown>(
...
       const contractInstance = contractFactory.at(previousDeployResult.contractInstance.address)
       return {
         ...previousDeployResult,
         contractInstance
       }
     }
-    console.log(`Deploying contract ${taskId}`)
-    console.log(`Deployer - group ${signer.group} - ${signer.address}`)
+    deploymentLogger(`Deploying contract ${taskId}`)
+    debugLogger(`Deployer - group ${signer.group} - ${signer.address}`)
```
Lbqds, line 399: We only log the deployer info in debug mode.

h0ngcha0 (approved):
LGTM :+1:

packages/cli/src/deployment.ts
```diff
@@ -323,7 +323,14 @@ function getTaskId(code: Contract | Script, taskTag?: string): string {
...
 }
 
+function debugLog(enableDebugMode: boolean, message: string) {
+  if (enableDebugMode) {
+    console.log(message)
+  }
+}
+
 function createDeployer<Settings = unknown>(
+  enableDebugMode: boolean,
```
polarker: `debugModeEnabled: boolean`

polarker (approved):
LGTM

PR #400, opened 2024-07-23
Support calling method multiple times in milticall

Lbqds:


artifacts/ts/Add.ts
```diff
@@ -391,4 +433,15 @@ export class AddInstance extends ContractInstance {
       getContractByCodeHash
     )) as AddTypes.MultiCallResults<Calls>;
   }
+
+  async multicallByMethod<Calls extends AddTypes.MultiCallByMethodParams>(
+    calls: Calls
+  ): Promise<AddTypes.MultiCallByMethodResults<Calls>> {
```
Lbqds, line 439: Because the parameter is an array, typescript cannot accurately infer the return type. The return type would be `(CallResult0 | CallResult1 | CallResult2)[]`, user needs to convert the type manually. So I kept the `multicall` method, it would be better to use `multicall` if the user only needs to call a method once.

polarker (approved):
LGTM!

PR #401, opened 2024-07-25
Generate typescript code for global constants

Lbqds:


contracts/constants.ral
```diff
@@ -0,0 +1,26 @@
...
+enum Addresses {
+    A = @1DrDyTr9RpRsQnDnXo2YRiPzPW4ooHX5LLoqXrqfMrpQH
+    B = @14UAjZ3qcmEVKdTo84Kwf4RprTQi86w2TefnnGFjov9xF
+}
+
+enum ByteVecs {
+    A = #00
+    B = #11
+}
+
```
polarker, line 20: Let's test Boolean constants as well

packages/cli/src/codegen.ts
```diff
@@ -228,27 +228,31 @@ function genEventIndex(contract: Contract): string {
...
+  return constants.map((constant) => `${prefix}${constant.name}${separator}${nodeValToString(constant.value)}`)
 }
 
-function genEnum(enumDef: Enum): string {
+function genEnum(enumDef: Enum, isLocal: boolean): string {
+  const prefix = isLocal ? '' : 'export const '
+  const separator = isLocal ? ': ' : ' = '
   const fields = enumDef.fields.map((field) => `${field.name}: ${nodeValToString(field.value)}`).join(',')
-  return `${enumDef.name}: { ${fields} }`
+  return `${prefix}${enumDef.name}${separator}{ ${fields} }`
```
polarker, line 251: Let's use typescript enum: https://www.typescriptlang.org/docs/handbook/enums.html

Lbqds: We convert `U256` and `I256` to `bigint`, but typeScript does not support `bigint` as an enum constant: https://github.com/microsoft/TypeScript/issues/37783

polarker: Interest. Let's use object then

polarker (approved):
LGTM

PR #403, opened 2024-07-29
Minor Updates

polarker:


packages/web3/src/contract/contract.ts
```diff
@@ -1725,14 +1725,15 @@ function getBytecodeTemplate(
   // For the default TxScript main function
   const numberOfMethods = '01'
   const isPublic = '01'
-  const modifier = usePreapprovedAssets ? '03' : '00'
+  const scriptUseApprovedAssets = attoAlphAmount !== undefined || tokens !== undefined
```
Lbqds, line 1728: Question: In what situations does a contract method use preapproved assets, but the script does not use preapproved assets?

Lbqds (approved):
LGTM, only one question.

packages/web3/src/contract/contract.ts
polarker: That's impossible, but it's possible that the method does not use preapproved assets, but the script use preapproved assets. E.g. the withdraw function of token faucet.

Lbqds: Got it, thanks for the explanation

PR #405, opened 2024-08-06
Remove binary-parser and refactor codec

Lbqds:


packages/web3/src/codec/bigint-codec.ts
polarker, line 1: Let's refactor this to be a real decoder instead of relying on the signed parameter.

packages/web3/src/codec/contract-output-ref-codec.ts
```diff
@@ -15,28 +15,17 @@ GNU Lesser General Public License for more details.
...
-  }
-
-  decode(input: Uint8Array): ContractOutputRef {
-    return this.parser.parse(input)
-  }
-}
-
-export const contractOutputRefCodec = new ContractOutputRefCodec()
+export const contractOutputRefCodec = new ObjectCodec<ContractOutputRef>({
+  hint: signedIntCodec,
```
polarker, line 28: Let's rename `signedIntCodec` as `intAs4BytesCodec`, since there is a compact integer codec for signed int.

packages/web3/src/codec/either-codec.ts
```diff
@@ -15,28 +15,18 @@ GNU Lesser General Public License for more details.
...
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
-import { Parser } from 'binary-parser'
 import { Codec } from './codec'
+import { Reader } from './reader'
 
 export interface Either<L, R> {
   either: number
   value: L | R
 }
```
polarker, line 24: Let's refactor this to use enum

`{ kind: Left, value: L } | { kind: Right, value: R }`

packages/web3/src/codec/instr-codec.ts
polarker, line 1: Let's generate this with a Scala tool? Btw, we should merge your Rust tool for generating the Rust codec.

packages/web3/src/codec/long-codec.ts
```diff
@@ -15,14 +15,10 @@ GNU Lesser General Public License for more details.
...
-import { Parser } from 'binary-parser'
 import { Codec, assert } from './codec'
+import { Reader } from './reader'
 
-export class LongCodec implements Codec<bigint> {
-  parser = Parser.start().buffer('value', {
-    length: 8
-  })
-
+export class LongCodec extends Codec<bigint> {
```
polarker, line 21: Do we use this fixed 8-byte codec anywhere? If not, we could remove it.

packages/web3/src/codec/val.ts
```diff
@@ -0,0 +1,39 @@
...
+import { boolCodec, EnumCodec } from './codec'
+import { LockupScript, lockupScriptCodec } from './lockup-script-codec'
+import { ArrayCodec } from './array-codec'
+
+export type Val =
+  | { type: 'Bool'; value: boolean }
+  | { type: 'I256'; value: bigint }
+  | { type: 'U256'; value: bigint }
+  | { type: 'ByteVec'; value: ByteString }
+  | { type: 'Address'; value: LockupScript }
```
polarker, line 30: Let's use `kind` instead of `type`

polarker (commented):
Great improvements 👍 I added a few more TODO improvements.

packages/web3/src/codec/instr-codec.ts
Lbqds: Will add a CLI tool 👍

polarker (approved):
LGTM 🚀

PR #407, opened 2024-08-13
Simple script vm

polarker:
A symbolic execution vm for tx scripts. Most of the TxScript instructions are implemented, we could add more based on feedback.

Not many unit tests for now as I don't have much time. We could add them in the future.


packages/web3/src/contract/script-simulator.ts
```diff
@@ -0,0 +1,837 @@
...
+type SymbolI256 = { kind: 'Symbol-I256'; value: undefined }
+type SymbolU256 = { kind: 'Symbol-U256'; value: undefined }
+type SymbolByteVec = { kind: 'Symbol-ByteVec'; value: undefined }
+type SymbolAddress = { kind: 'Symbol-Address'; value: undefined }
+type SymbolAny = { kind: 'Symbol-Any'; value: undefined }
+type SimulatorVal = Val | SymbolBool | SymbolI256 | SymbolU256 | SymbolByteVec | SymbolAddress | SymbolAny
+type SimulatorVar<K extends string, V> = { kind: K; value: V } | { kind: `Symbol-${K}`; value: undefined }
+
+function unaryOp<K extends string, V>(x: SimulatorVar<K, V>, op: (x: V) => V): SimulatorVar<K, V> {
+  if (x.kind.startsWith('symbol')) {
```
Lbqds, line 595: Do we need to use `Symbol` instead of `symbol` in these `*Op` functions? It seems that `'Symbol-Bool'.startsWith('symbol')` returns false.

test/transact.test.ts
```diff
@@ -16,39 +16,242 @@ You should have received a copy of the GNU Lesser General Public License
...
+    expect(getCalls0[0].approvedTokens).toBe(undefined)
+    const getTx1 = await instance.transact.getTotalALPH({ signer, attoAlphAmount: ONE_ALPH })
+    const getCalls1 = ScriptSimulator.extractContractCallsWithErrors(getTx1.unsignedTx)
+    expect(getCalls1.length).toBe(1)
+    expect(getCalls1[0].contractAddress).toBe(instance.address)
+    expect(getCalls1[0].approvedAttoAlphAmount).toBe(undefined)
+    expect(getCalls1[0].approvedTokens).toBe(undefined)
+  })
+
+  it.only('should use transact methods for token', async function () {
```
Lbqds, line 88: Do we need to remove the `only` here?

test/transact.test.ts
```diff
@@ -16,39 +16,242 @@ You should have received a copy of the GNU Lesser General Public License
...
+      initialAttoAlphAmount: ONE_ALPH,
+      initialFields: { tokenId, totalALPH: 0n, totalTokens: 0n },
+      issueTokenAmount: 10n ** 18n,
+      issueTokenTo: signer.address
+    })
+    const deployCalls2 = ScriptSimulator.extractContractCallsWithErrors(deploy2.unsignedTx)
+    expect(deployCalls2.length).toBe(0)
+  })
+
+  it.only('should test multicall', async function () {
```
Lbqds, line 197: Same here.

packages/web3/src/contract/script-simulator.ts
```diff
@@ -506,14 +527,24 @@ export class ScriptSimulator {
...
         }
         case 'ApproveAlph': {
-          operandStack.popU256() // amount
-          operandStack.popAddress() // spender
+          const amount = operandStack.popU256() // amount
+          const spender = operandStack.popAddress() // spender
+          if (spender.kind.startsWith('Symbol')) {
+            approved.setUnknown() // The spender might be the caller
+          } else if (spender === callerAddress) {
+            approved.addApprovedAttoAlph(amount)
```
h0ngcha0, line 535: Shall we set the approved to `unknown` as well if `callerAddress` isn't spender?

polarker: We only show the approvals for call address in the return, so we don't need to care about other addresses.

polarker:
@Lbqds @h0ngcha0 I have resolved all comments and will release a new version. If you have more comments, please feel free to add them here.

packages/web3/src/contract/script-simulator.ts
h0ngcha0: I am thinking about perhaps this would underestimate the total amount of approvals in corner cases, since we could have something like

```
foo.call{callerAddress!() -> 1 alph, alice -> 2 alph)
```

where alice is actually the same as callerAddress!(). Will this happen?

polarker: If alice's address is unknown, then branch `spender.kind.startsWith('Symbol')` will be executed, the approved assets would be set to unknown.

If alice's address is known, then we will compare if the address is equal to caller address.

PR #411, opened 2024-08-23
Fix loading compiler options

Lbqds:
Because we've introduced a new compiler option, an error will be thrown when loading incomplete compiler options. This will cause the project before the Rhone upgrade to be recompiled. This PR resolves this issue.


polarker:
@Lbqds Please make a new release for me

Issue #412, opened 2024-08-28
Don't show desktop wallet connect when the WC provider is on mobile devices


polarker:
<img width="607" alt="Screenshot 2024-08-28 at 17 32 22" src="https://github.com/user-attachments/assets/ff6acced-1754-47f4-aa32-e378efd6f034">

PR #413, opened 2024-08-29
Don't display the desktop wallet connect option when using a mobile d…

Lbqds:
Resolves issue #412

### DIFF
--- a/packages/web3-react/src/components/Pages/Connectors/index.tsx
+++ b/packages/web3-react/src/components/Pages/Connectors/index.tsx
@@ -15,7 +15,7 @@ GNU Lesser General Public License for more details.
 You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
-import React from 'react'
+import React, { useMemo } from 'react'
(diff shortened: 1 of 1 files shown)


nop33:
Now that we are at it, shouldn't we also hide the "Extension Wallet" when we are on mobile? What's the point of showing it there? That would mean that when we are on mobile, clicking the "Connect" button will only show 1 connector: WalletConnect. Couldn't we simplify that further and directly trigger the WalletConnect button when the "Connect" button is clicked? This saves the mobile user some confusion and some clicks.

cc @polarker @Lbqds

Lbqds:
> Now that we are at it, shouldn't we also hide the "Extension Wallet" when we are on mobile? What's the point of showing it there? That would mean that when we are on mobile, clicking the "Connect" button will only show 1 connector: WalletConnect. Couldn't we simplify that further and directly trigger the WalletConnect button when the "Connect" button is clicked? This saves the mobile user some confusion and some clicks.
> 
> cc @polarker @Lbqds

Sounds great 👍, I will improve this.

MaudSimon:
I just want to mention that there are users using the web extension on mobile using browsers like Mises browser.

PR #416, opened 2024-09-04
Fix audit with `pnpm audit --prod --fix`

nop33:


.github/workflows/audit.yml
```diff
@@ -22,4 +22,4 @@ jobs:
       - uses: pnpm/action-setup@v4
         with:
           version: 9.4
-      - run: pnpm audit
+      - run: pnpm audit --prod
```
nop33, line 25: @h0ngcha0 don't you think it makes sense to only audit packages that are used in production

https://pnpm.io/cli/audit#--prod--p

h0ngcha0: Sorry for being late, would it be an issue if we audit all packages? `devDependencies` might affect our dev environment so if we can maybe we should audit as well, WDYT?

nop33: ok, fair enough!

polarker:
@nop33 It seems only webpack needs to be patched for dev deps. Can you give it a try?

PR #418, opened 2024-09-05
Return an array even if the result of multiple calls has only one element

### LINKED ISSUE #417: Roll back the return types of multicall
Let's use an array for returns of multicall even if the number of call objects is just 1, which is more consistent.

Since multicall is an advanced feature used by devs, there is no need to optimize the return types.

Lbqds:
Resolves #417


polarker (approved):
LGTM

PR #420, opened 2024-09-10
Support testing contract private functions in integration tests

Lbqds:
If `exposePrivateFunctions` is true, all contract functions will be changed to public and then the modified contract will be deployed. The downside is that we have to generate code for the contract's private functions.

### DIFF
--- a/packages/web3/src/contract/contract.ts
+++ b/packages/web3/src/contract/contract.ts
@@ -197,7 +197,7 @@ export class Contract extends Artifact {
 
   readonly bytecodeDebug: string
   readonly codeHashDebug: string
-  readonly decodedDebugContract: contract.Contract
+  readonly decodedContract: contract.Contract
 
   private bytecodeForTesting: string | undefined
   private codeHashForTesting: string | undefined
@@ -233,7 +233,7 @@ export class Contract extends Artifact {
     this.bytecodeDebug = ralph.buildDebugBytecode(this.bytecode, this.bytecodeDebugPatch)
     this.codeHashDebug = codeHashDebug
 
-    this.decodedDebugContract = contract.contractCodec.decodeContract(hexToBinUnsafe(this.bytecodeDebug))
+    this.decodedContract = contract.contractCodec.decodeContract(hexToBinUnsafe(this.bytecode))
     this.bytecodeForTesting = undefined
     this.codeHashForTesting = undefined
   }
@@ -247,9 +247,10 @@ export class Contract extends Artifact {
       return this.bytecodeForTesting
     }
 
-    const methods = this.decodedDebugContract.methods.map((method) => ({ ...method, isPublic: true }))
+    const decodedDebugContract = contract.contractCodec.decodeContract(hexToBinUnsafe(this.bytecodeDebug))
+    const methods = decodedDebugContract.methods.map((method) => ({ ...method, isPublic: true }))
     const bytecodeForTesting = contract.contractCodec.encodeContract({
-      fieldLength: this.decodedDebugContract.fieldLength,
+      fieldLength: decodedDebugContract.fieldLength,
       methods: methods
     })
     const codeHashForTesting = blake.blake2b(bytecodeForTesting, undefined, 32)
@@ -263,7 +264,7 @@ export class Contract extends Artifact {
   }
 
   getDecodedMethod(methodIndex: number): Method {
-    return this.decodedDebugContract.methods[`${methodIndex}`]
+    return this.decodedContract.methods[`${methodIndex}`]
   }
 
   publicFunctions(): FunctionSig[] {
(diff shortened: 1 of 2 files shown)


h0ngcha0 (approved):
LGTM :+1:

packages/web3/src/contract/contract.ts
```diff
@@ -230,19 +233,53 @@ export class Contract extends Artifact {
...
+    this.codeHashForTesting = binToHex(codeHashForTesting)
+    return this.bytecodeForTesting
+  }
+
+  hasCodeHash(hash: string): boolean {
+    return this.codeHash === hash || this.codeHashDebug === hash || this.codeHashForTesting === hash
+  }
+
+  getDecodedMethod(methodIndex: number): Method {
+    return this.decodedDebugContract.methods[`${methodIndex}`]
```
polarker, line 266: Is `getDecodedMethod` used only for testing purpose? It always uses the debug contracts

test/contract.test.ts
```diff
@@ -761,4 +797,35 @@ describe('contract', function () {
     })
     expect(BigInt(state2.asset.alphAmount)).toEqual(MINIMAL_CONTRACT_DEPOSIT)
   })
+
+  it('should get the contract bytecode for testing', async () => {
+    expect(Add.contract.publicFunctions().length).not.toEqual(Add.contract.functions.length)
+    const instance0 = (await Add.deploy(signer, { initialFields: { sub: randomContractId(), result: 0n } }, true))
```
polarker, line 803: It's more readable if we put `exposePrivateFunctions` in params:
```
{ initialFields: { sub: randomContractId(), result: 0n }, exposePrivateFunctions: true }
```

test/contract.test.ts
```diff
@@ -761,4 +797,35 @@ describe('contract', function () {
...
+      .contractInstance
+    const state0 = await instance0.fetchState()
+    expect(state0.bytecode).toEqual(Add.contract.getByteCodeForTesting())
+    expect(state0.bytecode).not.toEqual(Add.contract.bytecode)
+    expect(state0.bytecode).not.toEqual(Add.contract.bytecodeDebug)
+    expect(state0.codeHash).not.toEqual(Add.contract.codeHash)
+    expect(state0.codeHash).not.toEqual(Add.contract.codeHashDebug)
+    expect(Add.contract.hasCodeHash(state0.codeHash)).toEqual(true)
+
+    expect(Assert.contract.publicFunctions().length).toEqual(Assert.contract.functions.length)
```
polarker, line 813: This should return the correct public function length instead of the length for testing contracts

packages/web3/src/contract/contract.ts
```diff
@@ -230,19 +233,53 @@ export class Contract extends Artifact {
...
+    return this.functions.filter((_, index) => this.getDecodedMethod(index).usePreapprovedAssets)
   }
 
   usingAssetsInContractFunctions(): FunctionSig[] {
-    return this.functions.filter((_, index) => this.decodedMethods[`${index}`].useContractAssets)
+    return this.functions.filter((_, index) => this.getDecodedMethod(index).useContractAssets)
+  }
+
+  isMethodUsePreapprovedAssets(methodIndex: number): boolean {
+    return this.getDecodedMethod(methodIndex).usePreapprovedAssets
```
polarker, line 282: These 3 functions can be used in non-testing environments. It's better not to use testing bytecode.

packages/web3/src/contract/contract.ts
```diff
@@ -230,19 +233,53 @@ export class Contract extends Artifact {
...
+  getByteCodeForTesting(): string {
+    if (this.bytecodeForTesting !== undefined) return this.bytecodeForTesting
+
+    if (this.publicFunctions().length == this.functions.length) {
+      this.bytecodeForTesting = this.bytecodeDebug
+      this.codeHashForTesting = this.codeHashDebug
+      return this.bytecodeForTesting
+    }
+
+    const methods = this.decodedDebugContract.methods.map((method) => ({ ...method, isPublic: true }))
```
polarker, line 250: Does this mean the bytecode for testing would expose private functions always, even if exposePrivateFunctions is false?

packages/web3/src/contract/contract.ts
Lbqds: The `getDecodedMethod` is only used for getting function modifiers. I fixed it in the latest commit.

test/contract.test.ts
Lbqds: We generate the testing bytecode only if the `exposePrivateFunctions` is true, and if all functions are public, we will use the `bytecodeDebug` directly. The purpose of this line of code is to check all functions are public.

packages/web3/src/contract/contract.ts
Lbqds: No, this function will be called only if the `exposePrivateFunctions` is true.

PR #422, opened 2024-09-10
Decode Unsigned Tx with Codec

h0ngcha0:
Use `codec` to decode unsigned transaction when `buildUnsignedTx`


packages/web3/src/signer/tx-builder.ts
```diff
@@ -113,16 +115,16 @@ export abstract class TransactionBuilder {
...
     return {
-      fromGroup: decoded.fromGroup,
-      toGroup: decoded.toGroup,
+      fromGroup: fromGroup,
+      toGroup: toGroup,
       unsignedTx: params.unsignedTx,
-      txId: decoded.unsignedTx.txId,
-      gasAmount: decoded.unsignedTx.gasAmount,
-      gasPrice: fromApiNumber256(decoded.unsignedTx.gasPrice)
+      txId: UnsignedTxCodec.txId(decoded),
```
Lbqds, line 125: The `UnsignedTx.txId` will encode the tx. Maybe we can directly calculate the tx id using the raw tx bytes.

packages/web3/src/transaction/utils.test.ts
```diff
@@ -0,0 +1,126 @@
...
+import { web3 } from '../'
+import { PrivateKeyWallet } from '@alephium/web3-wallet'
+
+describe('transaction utils', () => {
+  beforeAll(() => {
+    web3.setCurrentNodeProvider('http://127.0.0.1:22973', undefined, fetch)
+  })
+
+  it('should calculate the group of transfer transaction', async () => {
+    const [signer1, signer2] = await getSigners(2)
```
Lbqds, line 35: The `getSigners` function returns group 0 accounts by default. Maybe we can test the case where `fromGroup` and `toGroup` are not the same.

Lbqds (approved):
LGTM 👍

packages/web3/src/utils/group.ts
polarker, line 1: This seems better to be in `address` module?

packages/web3/src/utils/group.ts
```diff
@@ -0,0 +1,38 @@
...
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { TOTAL_NUMBER_OF_GROUPS } from '../constants'
+import djb2 from './djb2'
+
+export function groupFromBytes(bytes: Uint8Array): number {
```
polarker, line 22: I think this one should be a private function in `address`, as it's only used for asset addresses. The hints for contracts are different.

We could also rename it to be more informative: `groupFromBytesForAssetAddress()`

packages/web3/src/signer/tx-builder.ts
h0ngcha0: Fixed, thanks :+1:

packages/web3/src/utils/group.ts
h0ngcha0: Fixed, thanks :+1:

PR #423, opened 2024-09-11
Fix the unit tests for the nested map call

Lbqds:
We still need to keep the `initialMaps` in the `TestContractParams` because we cannot put the contract state that needs to be tested into `existingContracts`.


polarker (approved):
Great work 👍 A few minor remarks

PR #424, opened 2024-09-20
Multi Group Transfer

pragmaxim:
Demonstration of the future node's multi group transfer and how it could simplify transfers on the Tipping example.

### DIFF
--- a/packages/web3/src/signer/signer.ts
+++ b/packages/web3/src/signer/signer.ts
@@ -65,6 +65,7 @@ export abstract class SignerProvider {
   }
 
   abstract signAndSubmitTransferTx(params: SignTransferTxParams): Promise<SignTransferTxResult>
+  abstract signAndSubmitMultiGroupTransferTx(params: SignTransferTxParams): Promise<SignTransferTxResult[]>
   abstract signAndSubmitDeployContractTx(params: SignDeployContractTxParams): Promise<SignDeployContractTxResult>
   abstract signAndSubmitExecuteScriptTx(params: SignExecuteScriptTxParams): Promise<SignExecuteScriptTxResult>
   abstract signAndSubmitUnsignedTx(params: SignUnsignedTxParams): Promise<SignUnsignedTxResult>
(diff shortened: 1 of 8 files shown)


packages/web3/src/signer/signer.ts
```diff
@@ -118,7 +118,7 @@ export abstract class SignerProviderSimple extends SignerProvider {
     return signResult
   }
 
-  protected abstract getPublicKey(address: string): Promise<string>
+  abstract getPublicKey(address: string): Promise<string>
```
pragmaxim, line 121: Please let me know if the `SignerProviderSimple` can have this method public. Not important much, it would be easier for having a single method for build+sign+submit with just a `PrivateKeyWallet`

pragmaxim:
Guys openapi tests and tests are failing because the `update-schema:alephium` script is pointing to `alephium/` repo where this does not exist yet.

pragmaxim:
@h0ngcha0 the `lending` and `transactions` multi-group tests started failing after merging the new features in master on error : 
```
Failed in validating tx `foo` due to NonExistInput: `bar` 
```
Tried to troubleshoot but it will be a longer search

test/lending.test.ts
```diff
@@ -0,0 +1,199 @@
...
+  return result
+}
+
+describe('lendingbot', function () {
+  it('should work', async function () {
+    const lendingBot = new LendingBot(new NodeProvider('http://127.0.0.1:22973'))
+
+    // each user will start with 1 ALPH
+    const users = ['user0', 'user1', 'user2']
+    const deposit = convertAlphAmountWithDecimals('1.0')!
```
h0ngcha0, line 143: [Nit] There is a constant we can use: `ONE_ALPH`.

h0ngcha0:
> @h0ngcha0 the `lending` and `transactions` multi-group tests started failing after merging the new features in `dev-alephium/master` on error :
> 
> ```
> Failed in validating tx `foo` due to NonExistInput: `bar` 
> ```
> 
> Tried to troubleshoot but it will be a longer search, please let me know if you can think of the reason 🙏

I think there is a race condition here due to

```typescript
 return await Promise.all(
   buildTxResults.map(async (tx) => {
     return await signer.signAndSubmitUnsignedTx({
       signerAddress: params.signerAddress,
       unsignedTx: tx.unsignedTx
     })
   })
 )
```

devnet mine one tx into block at a time, so if the 2nd tx gets in before the 1st one, it will report `NonExistInput` error. If we do something like:

```
    const results: SignTransferTxResult[] = []
    for (const tx of buildTxResults) {
      const result = await signer.signAndSubmitUnsignedTx({
        signerAddress: params.signerAddress,
        unsignedTx: tx.unsignedTx
      })
      results.push(result)
    }
    return results
```
it should work.

Lbqds (approved):
LGTM 👍

h0ngcha0 (approved):
LGTM :+1:

PR #427, opened 2024-10-03
Support Chained Transactions

h0ngcha0:


docker/docker-compose.yml
```diff
@@ -29,7 +29,7 @@ services:
         condition: service_healthy
 
   alephium:
-    image: alephium/alephium:v3.7.0
+    image: liuhongchao/dev-alephium:3.7.0_29-43e3ae8d
```
h0ngcha0, line 32: Will update after full node is released.

packages/web3/src/api/api-alephium.ts
```diff
@@ -354,7 +354,13 @@ export interface BuildSweepMultisig {
 }
 
 /** BuildTransaction */
-export interface BuildTransaction {
+export type BuildTransaction = DeployContract | ExecuteScript | Transfer1
```
h0ngcha0, line 357: The name `Transfer1` and `Transfer2`, etc is due to us using `BuildTransaction.Transfer`  and `BuildTransactionResult.Transfer` at the full node side. Maybe we should create unique names for them such as `BuildTransferTransaction` and `BuildTransferTransactionResult`, etc. WDYT?

packages/web3/src/contract/contract.ts
```diff
@@ -565,7 +567,7 @@ export class Contract extends Artifact {
   async txParamsForDeployment<P extends Fields>(
     signer: SignerProvider,
     params: DeployContractParams<P>
-  ): Promise<SignDeployContractTxParams> {
+  ): Promise<SignDeployContractChainedTxParams> {
```
h0ngcha0, line 570: Here we are returning `SignDeployContractChainedTxParams` which contains two more fields: `publicKey` and `type`, which should be backwards compatible. 
Alternatively we can create another function to return it.

packages/web3/src/signer/tx-builder.ts
```diff
@@ -101,21 +90,71 @@ export abstract class TransactionBuilder {
...
+        }
+        case 'ExecuteScript': {
+          const buildExecuteScriptTxResult = buildResult.value
+          return {
+            ...buildExecuteScriptTxResult,
+            groupIndex: buildExecuteScriptTxResult.fromGroup,
+            gasPrice: fromApiNumber256(buildExecuteScriptTxResult.gasPrice),
+            type: buildResult.type
+          } as SignExecuteScriptChainedTxResult
+        }
```
polarker, line 146: Should we abstract out the conversion functions for TxResult?

packages/web3/src/signer/tx-builder.ts
```diff
@@ -62,28 +68,126 @@ export abstract class TransactionBuilder {
...
+  async buildExecuteScriptTx(
+    params: SignExecuteScriptTxParams,
+    publicKey: string
+  ): Promise<Omit<SignExecuteScriptTxResult, 'signature'>> {
+    const data = this.buildExecuteScriptTxParams(params, publicKey)
+    const response = await this.nodeProvider.contracts.postContractsUnsignedTxExecuteScript(data)
+    return { ...response, groupIndex: response.fromGroup, gasPrice: fromApiNumber256(response.gasPrice) }
+  }
+
+  async buildChainedTx(params: BuildChainedTxParams[]): Promise<BuildChainedTxResult[]> {
```
polarker, line 95: Probably more consistent and convenient to provide a list of public keys than to moving public keys to params.

```
async buildChainedTx(params: BuildChainedTxParams[], publicKeys: string[]): ...
```

packages/web3/src/signer/tx-builder.ts
```diff
@@ -62,28 +68,126 @@ export abstract class TransactionBuilder {
...
+        case 'DeployContract': {
+          const value = this.buildDeployContractTxParams(param, param.publicKey)
+          return { type: param.type, value }
+        }
+        case 'ExecuteScript': {
+          const value = this.buildExecuteScriptTxParams(param, param.publicKey)
+          return { type: param.type, value }
+        }
+        default:
+          throw new Error(`Unsupported transaction type: ${(param as any).type}`)
```
pragmaxim, line 111: `(param as any).type` is a bit suspicious as we do `switch (param.type)` a couple of lines above and bellow `${buildResult.type}`

h0ngcha0: Seems that typescript doesn't allow us to do `param.type` since it is already exhaustive (type-wise) for the switch statement, so it thinks it will not end up here.

h0ngcha0: Will get the value before switch statement to get rid of `as any`, thanks :+1:

packages/web3/src/contract/contract.ts
polarker: Since we decoupled the public key from tx parameters, can we keep the return type unchanged? Looks like possible.

Public keys can be fetched in other ways if necessary.

polarker: Even if I revert this to the original types without `type` field in the return, `npm run build` would still succeed. It seems there are some type safety issues here.

packages/web3/src/signer/signer.ts
```diff
@@ -40,10 +40,13 @@ import {
...
   SubmitTransactionParams,
   KeyType,
-  MessageHasher
+  MessageHasher,
+  SignChainedTxParams,
+  SignChainedTxResult
 } from './types'
 import { TransactionBuilder } from './tx-builder'
 import { addressFromPublicKey, groupOfAddress } from '../address'
+import { BuildChainedTx, BuildChainedTxResult } from '../api/api-alephium'
```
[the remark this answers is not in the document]
h0ngcha0: Fixed, thanks.

packages/web3-test/src/test-wallet.ts
```diff
@@ -96,8 +96,10 @@ export async function getSigner(alphAmount = ONE_ALPH * 100n, group = 0): Promis
     }
     const rootWallet = new PrivateKeyWallet({ privateKey: testPrivateKey })
     const wallet = PrivateKeyWallet.Random(group)
-    const destinations = [{ address: wallet.address, attoAlphAmount: alphAmount }]
-    await rootWallet.signAndSubmitTransferTx({ signerAddress: testAddress, destinations })
+    if (alphAmount > 0n) {
+      const destinations = [{ address: wallet.address, attoAlphAmount: alphAmount }]
+      await rootWallet.signAndSubmitTransferTx({ signerAddress: testAddress, destinations })
+    }
     return wallet
```
pragmaxim, line 103: I cannot make sense of it as a newcomer, we sign the transfer with `rootWallet` and return `PrivateKeyWallet.Random(group)` which seems to not hold any important information :thinking:

pragmaxim: I was writing similar tests last week while using this `getSigner()` method and the behavior was expected only when I returned and signed the transactions with the `rootWallet` :thinking:

packages/web3-test/src/test-wallet.ts
```diff
@@ -96,8 +96,10 @@ export async function getSigner(alphAmount = ONE_ALPH * 100n, group = 0): Promis
...
   } catch (_) {
     throw new Error('Failed to get signer, please restart the devnet')
```
pragmaxim, line 105: Please print the error as it reveals the reason of failure :
```
  } catch (error) {
    throw new Error(`Failed to get signer, please restart the devnet due to : ${error}`)
  }
```

polarker:
@Lbqds Please give a final review on this new feature, thanks!

Lbqds (approved):
LGTM 👍

PR #433, opened 2024-10-17
Improve the error messages

### LINKED ISSUE #429: Throw traceble error
We could use `throw new TraceableError ("Error message", err)` to include the stack trace
```
class TraceableError extends Error {
  trace: Error;

  constructor(message?: string, innerError?: Error) {
    super(message); 
    this.trace = innerError;

    const actualProto = new.target.prototype;

    if (Object.setPrototypeOf) { Object.setPrototypeOf(this, actualProto); } 
    else { this.__proto__ = actualProto; } 
  }
}
```

Lbqds:
Resolves #429


packages/cli/scripts/deploy.ts
```diff
@@ -35,8 +35,8 @@ export async function deployAndSaveProgress<Settings = unknown>(
     scriptExecuted = await deploy(configuration, networkId, deployments, fromIndex, toIndex, silent)
   } catch (error) {
     await deployments.saveToFile(deploymentsFile, configuration, false)
-    console.error(`Failed to deploy the project`)
-    throw error
+    console.error(`Failed to deploy the project, error: `, error)
```
polarker, line 38: Let's only log this when debug mod is enabled, since we throw error with stack trace now.

polarker (approved):
LGTM, just one remark.

PR #434, opened 2024-10-17
Print debug messages in integration tests

Lbqds:


packages/web3/src/contract/contract.ts
```diff
@@ -1876,7 +1876,11 @@ export async function signExecuteMethod<I extends ContractInstance, F extends Fi
     gasPrice: params.gasPrice
   }
 
-  return await signer.signAndSubmitExecuteScriptTx(signerParams)
+  const result = await signer.signAndSubmitExecuteScriptTx(signerParams)
+  if (await contract.contract.isDevnet(signer)) {
```
polarker, line 1880: It seems better to check `isContractDebugMessageEnabled` here, so it's more efficient if it's disabled.

PR #435, opened 2024-10-21
Esm support

polarker:


packages/web3-test/package.json
```diff
@@ -22,11 +16,25 @@
     "url": "https://github.com/alephium/alephium-web3/issues"
   },
   "author": "Alephium dev <dev@alephium.org>",
+  "main": "dist/_cjs/index.js",
+  "module": "dist/_esm/index.js",
+  "browser": "dist/alephium-web3-wallet.min.js",
```
nop33, line 21: This is the package.json of the web3-test package but web3-wallet is referenced in this line. Is this correct?

packages/web3-wallet/webpack.config.cjs
```diff
@@ -41,12 +41,15 @@ module.exports = {
...
   output: {
-    filename: 'alephium-web3-wallet.min.js',
+    filename: '_umd/alephium-web3-wallet.min.js',
     library: {
       name: 'alephium',
       type: 'umd'
     }
   },
+  experiments: {
+    asyncWebAssembly: true
```
Lbqds, line 51: Question: why do we need to add this to the `web3-wallet` package?

packages/web3-test/package.json
polarker: It's removed since there is not build for browser for this package.

packages/web3-wallet/webpack.config.cjs
polarker: I am not sure if this is still necessary, but I copied it from `web3`. IIRC, we needed it in the past for some project.

nop33:
Superseded by:
- https://github.com/alephium/alephium-web3/pull/544
Closing

PR #436, opened 2024-10-21
Websockets on the http rest port

pragmaxim:
This is a simple WS client that I was testing the WS server with from outside to be sure everything really works as expected.


package.json
```diff
@@ -5,12 +5,14 @@
     "@alephium/cli": "workspace:*",
     "@alephium/web3": "workspace:*",
     "@alephium/web3-test": "workspace:*",
-    "@alephium/web3-wallet": "workspace:*"
+    "@alephium/web3-wallet": "workspace:*",
+    "ws": "^8.5.4"
```
pragmaxim, line 9: The `ws` deps should be in `web3` however can't make it work this way, `js` newbie

PR #441, opened 2024-11-04
Remove the @alephium/cli dependency from the generated code

### LINKED ISSUE #440: Deployment artifacts depends on cli package
Let's make all artifacts depend only on the core web3 package.

Lbqds:
Resolves #440


polarker (approved):
LGTM

PR #448, opened 2024-11-16
Rename ZERO_ADDRESS to NULL_CONTRACT_ADDRESS

h0ngcha0:
Make it consistent with `nullContractAddress` in Ralph.


packages/web3/src/constants.ts
```diff
@@ -22,7 +22,7 @@ export const MIN_UTXO_SET_AMOUNT = BigInt(1000000000000)
 export const ALPH_TOKEN_ID = ''.padStart(64, '0')
 export const ONE_ALPH = 10n ** 18n
 export const DUST_AMOUNT = 10n ** 15n
-export const ZERO_ADDRESS = 'tgx7VNFoP9DJiFMFgXXtafQZkUvyEdDHT9ryamHJYrjq'
+export const NULL_CONTRACT_ADDRESS = 'tgx7VNFoP9DJiFMFgXXtafQZkUvyEdDHT9ryamHJYrjq'
```
Lbqds, line 25: Perhaps we could add a new `NULL_CONTRACT_ADDRESS` constant, which would avoid breaking compatibility with existing code, WDYT?

packages/web3/src/constants.ts
```diff
@@ -23,6 +23,7 @@ export const ALPH_TOKEN_ID = ''.padStart(64, '0')
 export const ONE_ALPH = 10n ** 18n
 export const DUST_AMOUNT = 10n ** 15n
 export const ZERO_ADDRESS = 'tgx7VNFoP9DJiFMFgXXtafQZkUvyEdDHT9ryamHJYrjq'
```
polarker, line 25: Since we are going to use tsdoc, let's try to add a deprecate warning for `ZERO_ADDRESS`

polarker (approved):
LGTM

PR #449, opened 2024-11-18
Detect injected providers

Lbqds:
When both the onekey wallet and the alephium wallet are enabled(I can't connect to the onekey extension wallet because I don't have the onekey hardware wallet.):

https://github.com/user-attachments/assets/57947597-3485-46b1-84c4-9287f9470d1e

When only the alephium wallet is enabled

https://github.com/user-attachments/assets/73984ad4-73b4-4837-bbf2-2cfbecb8c4f8

When only the onekey wallet is enabled

https://github.com/user-attachments/assets/d8a0f1ac-423b-4f43-826f-8562c418b47c


packages/web3-react/src/utils/providers.ts
```diff
@@ -0,0 +1,76 @@
...
+  const addNewProvider = (provider: AlephiumWindowObject) => {
+    if (allProviders.find((p) => p.icon === provider.icon) === undefined) {
+      allProviders.push(provider)
+      listeners.forEach((listener) => listener([...allProviders]))
+    }
+  }
+
+  const detectProviders = () => {
+    if (typeof window !== 'undefined') {
+      const oneKeyProvider = window['alephium']
```
Lbqds, line 36: Currently, the onekey wallet does not announce the alephium provider, and since onekey sets `window.alephium`, we will try to load the onekey provider from `window.alephium`.

packages/web3-react/src/utils/providers.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (!!oneKeyProvider && isWalletObj(oneKeyProvider)) {
+        addNewProvider(oneKeyProvider)
+      }
+
+      const defaultProvider = getWalletObject(alephiumProvider.id)
```
Lbqds, line 41: This is to ensure compatibility with older versions of the alephium extension wallet. If only the alephium wallet is enabled and no provider is announced, we will directly try to load the provider from `window.alephiumProviders.alephium`.

packages/web3-react/src/utils/providers.ts
```diff
@@ -0,0 +1,76 @@
...
+import { alephiumProvider, AlephiumWindowObject, getWalletObject, isWalletObj } from '@alephium/get-extension-wallet'
+
+export type InjectedProviderListener = (providers: AlephiumWindowObject[]) => void
+
+function createProviderStore() {
+  const listeners: Set<InjectedProviderListener> = new Set()
+  let allProviders: AlephiumWindowObject[] = []
+
+  const addNewProvider = (provider: AlephiumWindowObject) => {
+    if (allProviders.find((p) => p.icon === provider.icon) === undefined) {
```
Lbqds, line 28: Currently, the onekey wallet and the alephium wallet use the same id and name, so we can only differentiate between them using the icon.

packages/web3-react/src/utils/connector.ts
```diff
@@ -139,17 +155,32 @@ const injectedDisconnect = async (signerProvider: SignerProvider): Promise<void>
...
+      networkId: options.network,
+      onDisconnected: undefined as any
+    }
+    for (const provider of allProviders) {
+      const enabledAccount = await provider.enableIfConnected(enableOptions as any)
+      if (enabledAccount) {
+        await options.onConnected({ account: enabledAccount, signerProvider: provider })
+        setLastConnectedAccount('injected', enabledAccount, options.network)
+        // eslint-disable-next-line
+        ;(provider as any)['onDisconnected'] = options.onDisconnected
```
Lbqds, line 179: This is to work around the `enableIfConnected` serialization issue in onekey wallet: https://github.com/OneKeyHQ/cross-inpage-provider/blob/23a0e8e3ef2ace42e8a0359be49ddfd2a653d281/packages/providers/onekey-alph-provider/src/OnekeyAlphProvider.ts#L147, it should handle the callback in the same way as `unsafeEnable`.

polarker (approved):
LGTM

Lbqds:
Please don't merge this PR. I tested with `react-dapp-template` and `nextjs-app-dapp-template` without any issues, but when testing with bridge ui, sometimes it renders multiple times, causing issues with wallet listing and reconnection. I will continue to investigate and resolve this issue.

packages/web3-react/src/hooks/useInjectedProviders.tsx
```diff
@@ -16,19 +16,12 @@ You should have received a copy of the GNU Lesser General Public License
...
-  useEffect(() => {
-    const cancel = injectedProviderStore.subscribe(setProviders)
-    return () => {
-      cancel()
-    }
-  }, [setProviders])
-
-  return providers
-}
+export const useInjectedProviders = () =>
```
Lbqds, line 22: Even though we check if the provider already exists when adding, sometimes duplicate providers appear when refreshing the page due to caching. This issue does not occur in `react-dapp-template` or `nextjs-app-dapp-template` but sometimes occurs in bridge ui.

In the latest commit, `useSyncExternalStore` will be used instead of `useEffect/useState`. The `useSyncExternalStore` hook was introduced in React 18, but to maintain compatibility with React 17, we need to import it from `use-sync-external-store/shim`, which is an official library provided by React.

packages/web3-react/src/components/AlephiumConnect.tsx
```diff
@@ -176,39 +178,57 @@ export const AlephiumConnectProvider: React.FC<{
...
+      updateAccount(undefined)
+      updateSignerProvider(undefined)
+    }
+    const onConnected = (result: ConnectResult) => {
+      updateAccount(result.account)
+      updateSignerProvider(result.signerProvider)
+    }
+
+    const reconnectByInjected = async (providers: AlephiumWindowObject[], times = 0) => {
+      if (times >= 20) {
```
Lbqds, line 192: The reason I made this change is that our extension wallet injects the inpage provider using [a <script> tag](https://github.com/alephium/extension-wallet/blob/3b679e5b674dc6b6de11147bc1326a3d48f9da4f/packages/extension/src/content.ts#L9), which may run asynchronously in chrome manifest v3 due to [this issue](https://issues.chromium.org/issues/41267807). As a result, the dApp sends a `requestProvider` request but may not immediately receive a response. I tested `react-dapp-template`, `nextjs-app-dapp-template`, and bridge ui. Only bridge ui has this issue.

In the ETH ecosystem, detecting injected providers works like this:

1. The extension wallet calls `window.addEventListener('requestProvider', () => window.dispatchEvent('announceProvider', provider))` to handle the dApp's `requestProvider` requests.
2. The dApp calls `window.dispatchEvent('requestProvider')` to request providers.

A crucial assumption is that step 1 must be completed before step 2 begins. Otherwise, the dApp may not receive the provider, and `wagmi` is based on this assumption.

I checked the OneKey extension wallet and the MetaMask extension wallet:

1. OneKey wallet places [`injected.js` in `content_scripts`](https://github.com/OneKeyHQ/app-monorepo/blob/503c27c66f5dc47ce56a8aeee1b709bf17e846d9/apps/ext/src/manifest/chrome_v3.js#L97). In `injected.js`, OneKey [registers the wallet provider for each chain](https://github.com/OneKeyHQ/cross-inpage-provider/blob/c17620a963ca9d09d4b1c7d739411baf66db2b0f/packages/injected/src/injectedExtension.ts#L16), which ensures that `injected.js` completed before the dApp starts.
2. Just like our extension wallet, MetaMask used to inject the in-page provider using the `<script>` tag before. However, due to the chrome bug mentioned above, it now uses `chrome.scripting.registerContentScripts` instead in [this PR](https://github.com/MetaMask/metamask-extension/pull/15448).

I tried using the approach from OneKey and MetaMask to modify our extension wallet, but it didn't work as expected. I'm not sure why yet, perhaps it needs some other related code changes. So I've fixed the issue in the SDK, and it can work properly in different environments.

polarker: Could you check how this is resolved in Argent X's latest code?

Lbqds: I just checked the [starknetkit]()https://github.com/argentlabs/starknetkit code; unfortunately, they haven't addressed this issue. Starknetkit tries to load injected providers when the dApp starts, but due to the issue I described above, it sometimes fails to detect the installed Argent-X wallet. Here's a simple video demo where you can see that I have the Argent-X wallet installed, but when I refresh the page, there's a high chance that it will prompt me again to install the Argent-X wallet:

https://github.com/user-attachments/assets/dffbba8c-4a9e-4892-8adb-7479bdcf1f90

I'm not sure why no one has raised this issue with starknetkit. One possible reason is that they haven't tested with create-react-app. In the environments I've tested, this issue only occurs when using create-react-app. It never happens with vite or nextjs.

Lbqds: I tried using the MetaMask approach, and it works now. So we can remove the last commit of this PR.

Do you guys think we should fix this issue in the extension wallet? If so, I will test it in different browsers later and create a PR to update the extension wallet. cc @h0ngcha0

polarker: Yes, please fix the extension wallet accordingly! Thanks

polarker:
@Lbqds should we merge this now?

PR #450, opened 2024-11-19
Introduce DappTransactionBuilder

Lbqds:


polarker (approved):
LGTM 👍

packages/web3/src/contract/dapp-tx-builder.ts
```diff
@@ -0,0 +1,210 @@
...
+    tokens?: { id: HexString; amount: bigint }[]
+    retLength?: number
+  }) {
+    if (!isBase58(params.contractAddress)) {
+      throw new Error(`Invalid contract address: ${params.contractAddress}, expected a base58 string`)
+    }
+    const contractLockupScript = lockupScriptCodec.decode(base58ToBytes(params.contractAddress))
+    if (contractLockupScript.kind !== 'P2C') {
+      throw new Error(`Invalid contract address: ${params.contractAddress}, expected a P2C address`)
+    }
```
h0ngcha0, line 76: [Nit] Maybe we can check `contractAddress` using [isContractAddress](https://github.com/alephium/alephium-web3/blob/master/packages/web3/src/address/address.ts#L88)?

packages/web3/src/contract/dapp-tx-builder.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      BytesConst(hexToBinUnsafe('2233'))
+    ])
+
+    expect(genArgs([true, { array0: [1n, 2n], array1: [-1n, -2n] }])).toEqual([
+      ConstTrue,
+      U256Const1,
+      U256Const2,
+      I256ConstN1,
+      I256Const(-2n)
+    ])
```
h0ngcha0, line 62: [Nit] Can we test a case where arg is an object?

packages/web3/src/contract/dapp-tx-builder.ts
```diff
@@ -0,0 +1,210 @@
...
+export class DappTransactionBuilder {
+  private callerLockupScript: LockupScript
+  private approvedAssets: Map<HexString, bigint>
+  private instrs: Instr[]
+
+  constructor(public readonly callerAddress: string) {
+    try {
+      this.callerLockupScript = lockupScriptCodec.decode(base58ToBytes(this.callerAddress))
+      if (this.callerLockupScript.kind !== 'P2PKH' && this.callerLockupScript.kind !== 'P2SH') {
+        throw new Error(`Expected a P2PKH address or P2SH address`)
```
h0ngcha0, line 53: Q: Shall we allow P2SH?

h0ngcha0 (commented):
LGTM, only minor comments & questions

packages/web3/src/contract/dapp-tx-builder.test.ts
Lbqds: This test is used to test an object. The second element of the array here is an object.

packages/web3/src/contract/dapp-tx-builder.ts
Lbqds: Since both our private key wallet and the build-tx endpoint support the Schnorr key type, I think we should support this as well

h0ngcha0: That makes sense :+1:

PR #471, opened 2024-11-25
test: improved tests for `packages/web3/src/address`

### LINKED ISSUE #451: Improve Test Coverage for `packages/web3/src/address`
This issue aims to improve the test coverage for this folder by:  
1. Identifying untested or under-tested functions and classes.  
2. Writing additional unit tests to cover edge cases and scenarios.

SoarinSkySagar:
resolves #451

### DIFF
--- a/packages/web3/src/address/address.test.ts
+++ b/packages/web3/src/address/address.test.ts
@@ -17,194 +17,316 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
 import {
-  addressFromContractId,
-  addressFromPublicKey,
-  contractIdFromAddress,
-  groupOfAddress,
-  groupOfPrivateKey,
-  publicKeyFromPrivateKey,
-  subContractId,
-  tokenIdFromAddress,
   validateAddress,
+  isValidAddress,
   isAssetAddress,
   isContractAddress,
-  isValidAddress,
+  groupOfAddress,
+  contractIdFromAddress,
+  tokenIdFromAddress,
+  groupOfPrivateKey,
+  publicKeyFromPrivateKey,
+  addressFromPublicKey,
   addressFromScript,
+  addressFromContractId,
+  contractIdFromTx,
+  subContractId,
   groupOfLockupScript
 } from './address'
 import { binToHex, bs58 } from '../utils'
 import { randomBytes } from 'crypto'
-import { LockupScript, lockupScriptCodec } from '../codec/lockup-script-codec'
+import { LockupScript } from '../codec/lockup-script-codec'
 
-describe('address', function () {
-  it('should validate address', () => {
-    expect(validateAddress('15EM5rGtt7dPRZScE4Z9oL2EDfj84JnoSgq3NNgdcGFyu')).toBeUndefined()
-    expect(validateAddress('15EM5rGtt7dPRZScE4Z9oL2EDfj84JnoSgq3NNgdcGFy')).toBeUndefined()
-    expect(() => validateAddress('15EM5rGtt7dPRZScE4Z9oL2EDfj84JnoSgq3NNgdcGF')).toThrow('Invalid address:')
-    expect(
-      validateAddress('2jW1n2icPtc55Cdm8TF9FjGH681cWthsaZW3gaUFekFZepJoeyY3ZbY7y5SCtAjyCjLL24c4L2Vnfv3KDdAypCddfAY')
-    ).toBeUndefined()
-    expect(() =>
-      validateAddress('2jW1n2icPtc55Cdm8TF9FjGH681cWthsaZW3gaUFekFZepJoeyY3ZbY7y5SCtAjyCjLL24c4L2Vnfv3KDdAypCddfA')
-    ).toThrow('Invalid address:')
-    expect(validateAddress('eBrjfQNeyUCuxE4zpbfMZcbS3PuvbMJDQBCyk4HRHtX4')).toBeUndefined()
-    expect(() => validateAddress('eBrjfQNeyUCuxE4zpbfMZcbS3PuvbMJDQBCyk4HRHtX')).toThrow('Invalid address:')
-    expect(validateAddress('yya86C6UemCeLs5Ztwjcf2Mp2Kkt4mwzzRpBiG6qQ9kj')).toBeUndefined()
-    expect(() => validateAddress('yya86C6UemCeLs5Ztwjcf2Mp2Kkt4mwzzRpBiG6qQ9k')).toThrow('Invalid address:')
-    expect(() => validateAddress('asd')).toThrow('Invalid multisig address: asd')
-    expect(() => validateAddress('asdasdf')).toThrow('Invalid multisig address')
-    expect(() =>
-      validateAddress('2jW1n2icPtc55Cdm8TF9FjGH681cWthsaZW3gaUFekFZepJoeyY3ZbY7y5SCtAjyCjLL24c4L2Vnfv3KDdAypCddfAY1')
-    ).toThrow('Invalid address:')
-    // both n and m are 0
-    expect(() => validateAddress('LUw')).toThrow('Invalid multisig address')
-    expect(() =>
-      validateAddress('2jVWAcAPphJ8ueZNG1BPwbfPFjjbvorprceuqzgmJQ1ZRyELRpWgARvdB3T9trqpiJs7f4GkudPt6rQLnGbQYqq2NCi')
-    ).toThrow('Invalid multisig address, n: 2, m: 3')
-    expect(() => validateAddress('thebear')).toThrow('Invalid multisig address')
+describe('address', () => {
+  describe('validateAddress', () => {
+    it('should validate valid P2PKH addresses', () => {
+      expect(() => validateAddress('15EM5rGtt7dPRZScE4Z9oL2EDfj84JnoSgq3NNgdcGFyu')).not.toThrow()
+      expect(() => validateAddress('1D59jXR9NpD9ZQqZTRVcVbKVh6ko5TUMt89WvkA8P9P7w')).not.toThrow()
+    })
+
+    it('should validate valid P2MPKH addresses', () => {
+      expect(() =>
+        validateAddress(`'2jW1n2icPtc55Cdm8TF9FjGH681cWthsaZW3gaUFekFZepJoeyY3ZbY7y5SCtAjyCjLL24c4L2Vnfv3KDdAypCddfAY'`)
+      ).not.toThrow()
+    })
+
+    it('should validate valid P2SH addresses', () => {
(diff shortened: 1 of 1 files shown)


SoarinSkySagar:
@polarker @kfastov PTAL

SoarinSkySagar:
PTAL @h0ngcha0

packages/web3/src/address/address.test.ts
```diff
@@ -17,194 +17,316 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+    })
   })
 
-  it('should calculate the group of private key', () => {
-    expect(groupOfPrivateKey('a5c91afebe25e1644e9023e0d341ca713b59beca36e6635de37ef0c9c8689654')).toEqual(3)
-    expect(groupOfPrivateKey('82e1f50a8e372933ab8afc6362934b693e5aaa4ca308b0aac27fbe8755e0a3fa')).toEqual(2)
-    expect(groupOfPrivateKey('860ef6d3468e93aee997617e3595859fbab634111b88a79099a7e359f447cc2f')).toEqual(2)
-    expect(groupOfPrivateKey('b3bda653c36f99ad5a88401ea99cbc2a3c25e74d3df652f1545439ec929bcf31')).toEqual(0)
+  describe('addressFromPublicKey', () => {
+    it('should compute address from public key', () => {
```
h0ngcha0, line 246: I suggest using `PrivateKeyWallet.Random()` to create pubic/private key pair for all tests involving the use of public key and private key, so we don't need to hardcoded ones.

packages/web3/src/address/address.test.ts
```diff
@@ -17,194 +17,316 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+      )
+    })
+  })
+
+  describe('contractIdFromTx', () => {
+    it('should compute contract ID from txId and outputIndex', () => {
+      const txId = '0a38bc48fbb4300f1e305b201cd6129372d867122efb814d871d18c0bfe43b56'
+      const outputIndex = 0
+      const contractId = contractIdFromTx(txId, outputIndex)
+      expect(contractId).toBeTruthy()
```
h0ngcha0, line 306: This is not verifying the contract id is correct, only that it returns something. We might need to create a contract id in integration test and verify it it matches what's returned by `contractIdFromTx(txId, outputIndex)`

packages/web3/src/address/address.test.ts
```diff
@@ -17,194 +17,316 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+      const address = bs58.encode(bytes)
+      expect(() => validateAddress(address)).toThrow(`Invalid address: ${address}`)
+    })
+
+    it('should throw error for invalid multisig address', () => {
+      const invalidAddressType = 0x01
+      const invalidData = randomBytes(10)
+      const bytes = Buffer.concat([Buffer.from([invalidAddressType]), invalidData])
+      const address = bs58.encode(bytes)
+      expect(() => validateAddress(address)).toThrow(`Invalid multisig address: ${address}`)
```
h0ngcha0, line 82: The exception is not deterministic, sometimes it could return `"Invalid multisig address, n: 0, m: xxxxx"`

packages/web3/src/address/address.test.ts
```diff
@@ -17,194 +17,316 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
-    const parentContractId = '0a38bc48fbb4300f1e305b201cd6129372d867122efb814d871d18c0bfe43b56'
-    const pathInHex = '4f51cd1f0af97cf5ec9c7a3397eaeea549d55a93c216e54f2ab4a8cf29f6f865'
-    expect(() => subContractId(parentContractId, pathInHex, -1)).toThrow('Invalid group -1')
-    expect(() => subContractId(parentContractId, pathInHex, 4)).toThrow('Invalid group 4')
-    expect(() => subContractId('&&', pathInHex, 0)).toThrow(`Invalid parent contract ID: &&, expected hex string`)
-    expect(() => subContractId(parentContractId, '&&', 0)).toThrow(`Invalid path: &&, expected hex string`)
-    expect(subContractId(parentContractId, pathInHex, 0)).toBe(
-      '0e28f15ca290002c31d691aa008aa56ac12356b0380efb6c88fff929b6a26800'
-    )
-    expect(subContractId(parentContractId, pathInHex, 1)).toBe(
```
h0ngcha0, line 200: These 3 tests are removed, is it intentional?

packages/web3/src/address/address.test.ts
```diff
@@ -17,194 +17,320 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+      expect(contractId).toBeTruthy()
+    })
+
+    it('should throw error for invalid tx', () => {
+      expect(() => tokenIdFromAddress('eBrjfQNeyUCuxE4zpbfMZcbS3PuvbMJDQBCyk4HRHtX4')).toThrow(
+        'Invalid contract address type: 2'
+      )
+    })
+
+    it('should handle large outputIndex values', () => {
```
h0ngcha0, line 315: Same as https://github.com/alephium/alephium-web3/pull/471/files#r1864293992

SoarinSkySagar:
PTAL @h0ngcha0

packages/web3/src/address/address.test.ts
```diff
@@ -17,51 +17,124 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+          publicKeyHashes: [randomBytes(32), randomBytes(32)],
+          m: mNegative
+        }
+      })
+      const addressNegative = bs58.encode(bytesNegative)
+      expect(() => validateAddress(addressNegative)).toThrow(`Invalid multisig address, n: ${n}, m: ${mNegative}`)
+    })
+
+    it('should throw error for invalid P2MPKH address where n < m', () => {
+      const addressType = 0x01
```
h0ngcha0, line 128: We can also use `lockupScriptCodec.encode` in this test, similar to how it is done in the test above.

packages/web3/src/address/address.test.ts
```diff
@@ -77,134 +150,237 @@ describe('address', function () {
...
+    })
+  })
+
+  describe('isContractAddress', () => {
+    it('should return true for valid contract addresses', () => {
+      expect(isContractAddress('yya86C6UemCeLs5Ztwjcf2Mp2Kkt4mwzzRpBiG6qQ9kj')).toBe(true)
+    })
+
+    it('should return false for asset addresses', () => {
+      expect(isContractAddress('15EM5rGtt7dPRZScE4Z9oL2EDfj84JnoSgq3NNgdcGFyu')).toBe(false)
```
h0ngcha0, line 195: Maybe we can test P2SH and P2MPKH address too just like we did for the `should return true for valid asset addresses` tests.

packages/web3/src/address/address.test.ts
```diff
@@ -77,134 +150,237 @@ describe('address', function () {
...
-  })
-
-  it('should extract token id from addresses', () => {
-    expect(binToHex(tokenIdFromAddress('wCTeteGBeSEC54GpkS8jWBzYiYNTBUuTW3WzxGd9yExT'))).toBe(
-      '25469eb0d0d0a55deea832924547b7b166c70a3554fe321e81886d3c18f19d64'
-    ),
+        expect(
+          groupOfAddress('2jWukVCejM4Zifz9LvMG4dfR6SEecHLX8VqbswhGwnu61d28B861UhLu3ZmTHu4N14m1kk9rbxreBYzcxta1WPawKzG')
+        ).toBe(3),
+        expect(groupOfAddress('eBrjfQNeyUCuxE4zpbfMZcbS3PuvbMJDQBCyk4HRHtX4')).toBe(0),
```
h0ngcha0, line 224: The following seem to be P2SH and P2C addresses, they should belong to the tests below

packages/web3/src/address/address.test.ts
```diff
@@ -77,134 +150,237 @@ describe('address', function () {
...
+
+  describe('publicKeyFromPrivateKey', () => {
+    it('should compute public key from private key', () => {
+      expect(publicKeyFromPrivateKey('91411e484289ec7e8b3058697f53f9b26fa7305158b4ef1a81adfbabcf090e45')).toBe(
+        '030f9f042a9410969f1886f85fa20f6e43176ae23fc5e64db15b3767c84c5db2dc'
+      )
+    })
+  })
+
+  describe('publicKeyFromPrivateKey', () => {
```
h0ngcha0, line 291: This seems to be a duplicate to the previous test

packages/web3/src/address/address.test.ts
```diff
@@ -77,134 +150,237 @@ describe('address', function () {
...
+  })
+  describe('addressFromPublicKey', () => {
+    it('should compute address from public key', () => {
+      expect(addressFromPublicKey('030f9f042a9410969f1886f85fa20f6e43176ae23fc5e64db15b3767c84c5db2dc')).toBe(
+        '1ACCkgFfmTif46T3qK12znuWjb5Bk9jXpqaeWt2DXx8oc'
+      )
+    })
+
+    it('should throw error for invalid public key', () => {
+      expect(() => tokenIdFromAddress('eBrjfQNeyUCuxE4zpbfMZcbS3PuvbMJDQBCyk4HRHtX4')).toThrow(
```
h0ngcha0, line 306: Shall we test `addressFromPublicKey` function here?

h0ngcha0:
Please run `npm run lint -- --fix` to fix the formatting of the updated files.

pragmaxim:
Wouldn't it be a major improvement of these tests if the literal addresses and contractI-ids were extracted into variables of meaningful name, possibly generating them?

packages/web3/src/address/address.test.ts
SoarinSkySagar: removing this test, because the `addressFromPublicKey` does not have error catching logic, so all string will be accepted as valid pubkeys

packages/web3/src/address/address.test.ts
SoarinSkySagar: We are already doing that in `isAssetAddress` tests

SoarinSkySagar:
Made all the required changes according to your reviews @h0ngcha0 ser, PTAL

SoarinSkySagar:
> Please run `npm run lint -- --fix` to fix the formatting of the updated files.

Umm I don't think there are any formatting errors in the updated file

pragmaxim:
> I guess to generate the addresses I would need to import from @alephium/web3-wallet package, which is not possible as we're not able to import rn, as has been my conclusion from our chats on telegram group. So it has been done this way

Yes, but I noticed of same string literal to be used 5 times in the same spec which is a perfect candidate for a constant I think

SoarinSkySagar:
> Yes, but I noticed of same string literal to be used 5 times in the same spec which is a perfect candidate for a constant I think

Done, extracted all repetitive addresses and contractIds into separate variables

Issue #454, opened 2024-11-21
Improve Test Coverage for `packages/web3/src/signer`


polarker:
This issue aims to improve the test coverage for this folder by:  
1. Identifying untested or under-tested functions and classes.  
2. Writing additional unit tests to cover edge cases and scenarios.

khayss:
Hello.
I want to take up this issue.

## My Background
I'm a full-stack web3 dev. I have great experience writing tests for different projects and different languages.

## Action Plan
- Review the code base to get a good idea of all components.
- Review the test code to see areas not covered by current tests.
- Add more tests to cover this areas.

jaykayudo:
Let me handle this issue!

raizo07:
Can I tackle this one? I'm a Software developer with over four years of experience. 


I will identify the untested or under-tested functions and classes and write additional unit tests to cover edge cases and scenarios

NueloSE:
May I try my hand at this?

mimiprosper:
I'd like to take this issue. Am a web 2 frontend developer, a smart contract developer/auditor. Software Testing & security is an area am passionate about. I have participated successfully in OD Hack task assigned to me. I have done unit testing with mocha. I have done audits for lots of smart contract projects and found vulnerabilities in public & private project.

I would improve this code base test coverage by identifying tested, under-tested classes and functions. I would also write unit test to cover edge cases. I would conclude this task in 7 days if assigned this task.

No-bodyq:
May I try my hand at this?

Supa-mega:
Can I contribute to this one?

od-hunter:
May I handle this issue? My experience includes html, css, react, javaScript, typescript, solidity & Cairo and I'm good with testing.

I'll use a testing framework with a coverage tool to identify untested functions, methods, and classes then analyze the report for areas with low coverage, focusing on untested branches, conditions, or methods.
I'll identify critical untested functions, edge cases, and areas with complex logic, I'll focus on conditional branches, loops, and error-handling scenarios.
I'll create test cases targeting happy Path(Validate expected outcomes with valid inputs), Edge Cases(Test boundary inputs) and Error Scenarios( Validate handling of invalid inputs or exceptions).
And then test.

Kindly assign, I'm ready to work.

Kaminar-i:
Mind if I take this issue?
i'm new and would love to work on this

petersdt:
I’d like to work on this.

kfastov:
[...]

PR #474, opened 2024-11-26
test: Improve Test Coverage for packages/web3/src/signer

raizo07:
closes: #454

### DIFF
--- a/packages/web3/src/signer/types.test.ts
+++ b/packages/web3/src/signer/types.test.ts
@@ -1,0 +1,252 @@
+/*
+Copyright 2018 - 2022 The Alephium Authors
+This file is part of the alephium project.
+
+The library is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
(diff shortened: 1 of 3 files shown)


h0ngcha0:
Hey, please make sure that all the new tests can be compiled, executed and run successfully. Right now all of them can't be executed due to type errors.

packages/web3/src/signer/types.test.ts
```diff
@@ -0,0 +1,252 @@
...
+        expect(params.bytecode).toBe('0xabcdef')
+        expect(params.signerKeyType).toBe('default')
+        expect(params.attoAlphAmount).toBe('1000000000000000000')
+        expect(params.tokens).toHaveLength(1)
+        expect(params.gasAmount).toBe(150)
+        expect(params.gasPrice).toBe('1500000000000000')
+        expect(params.gasEstimationMultiplier).toBe(1.5)
+      })
+    })
+  })
```
pragmaxim, line 186: Most of these just test general javascript, it would just increase code coverage and maintenance costs I think.

raizo07: @h0ngcha0 okay will sort out the test cases

PR #477, opened 2024-12-02
Add browser entry in WC provider package.json

nop33:
Similarly to how the `@alephium/web3` package provides a [`"browser"` entry in its package.json](https://github.com/alephium/alephium-web3/blob/master/packages/web3/package.json#L7), so must the `@alephium/walletconnect-provider`. It helps frontend projects chose the right build file for their environment.


h0ngcha0 (approved):
LGTM :+1:

PR #480, opened 2024-12-05
Generate ralph interfaces based on contract artifacts

Lbqds:
Since the contract's function parameters can be contract, struct, or interface types, we need to generate the interface by parsing the entire artifacts directory.


packages/cli/src/gen-interfaces.ts
```diff
@@ -0,0 +1,165 @@
...
+        const newPath = path.join(dirPath, dirent.name)
+        await load(newPath)
+      }
+    }
+  }
+  await load(artifactDir)
+  return contracts
+}
+
+async function getContractFromArtifact(filePath: string, structs: Struct[]): Promise<Contract | undefined> {
```
polarker, line 151: 1. Since all files are generated and should not be modified, maybe we prefix the file name with`__`. E.g. `__IFoo.ral`
2. Let's also add a comment `/// Autogenerated file. Do not edit manually.`

PR #481, opened 2024-12-10
Support extracting token deposit info from txs

Lbqds:


packages/web3/src/exchange/exchange.ts
```diff
@@ -67,7 +58,54 @@ export function getALPHDepositInfo(tx: Transaction): { targetAddress: Address; d
...
+    }
+  }
+  return inputAddresses
+}
+
+export function isTokenTransferTx(tx: Transaction): boolean {
+  return isTransferTx(tx) && checkTokenOutput(tx)
+}
+
+export function getTokenDepositInfo(tx: Transaction): {
```
polarker, line 81: Does this generate information for ALPH if ALPH and tokens are sent in a single transaction?

packages/web3/src/exchange/exchange.ts
```diff
@@ -62,6 +62,7 @@ function getInputAddresses(tx: Transaction): Address[] {
   const inputAddresses: Address[] = []
   for (const input of tx.unsigned.inputs) {
     try {
+      if (input.unlockScript === '03') continue // SameAsPrevious
```
polarker, line 65: IIRC, we have a constant somewhere for `03`, should reuse that one.

packages/web3/src/exchange/exchange.ts
Lbqds: No, it only returns the token info.

packages/web3/src/exchange/exchange.ts
Lbqds: We didn't define encoded `SameAsPrevious` before, so I added a constant for it in the codec

packages/web3/src/exchange/exchange.ts
```diff
@@ -74,36 +79,44 @@ function getInputAddresses(tx: Transaction): Address[] {
...
-  depositAmount: bigint
-}[] {
-  if (!isTokenTransferTx(tx)) return []
+export interface DepositInfo {
+  alph: BaseDepositInfo[]
+  tokens: TokenDepositInfo[]
+}
+
+export function getDepositInfo(tx: Transaction): DepositInfo {
+  if (!isTransferTx) return { alph: [], tokens: [] }
```
polarker, line 92: It should be `isTransferTx(tx)`

PR #484, opened 2024-12-25
Support custom connectors in web3-react

Lbqds:
To address [this WalletConnect issue](https://github.com/WalletConnect/walletconnect-monorepo/issues/5542) in bridge ui, we need to support passing more `SignClient` options in web3-react. This PR follows a similar approach to `wagmi` and `connectkit` by introducing a `connectors` parameter. In most cases, dApp developers do not need to specify this parameter.


polarker (approved):
LGTM 👍

PR #485, opened 2024-12-25
Use custom storage by default in the WalletConnect provider

Lbqds:


packages/walletconnect/src/provider.ts
```diff
@@ -333,7 +337,8 @@ export class WalletConnectProvider extends SignerProvider {
       (await SignClient.init({
         ...this.providerOpts,
         logger: this.providerOpts.logger || LOGGER,
-        relayUrl: this.providerOpts.relayUrl || RELAY_URL
+        relayUrl: this.providerOpts.relayUrl || RELAY_URL,
+        customStoragePrefix: this.getCustomStoragePrefix()
```
nop33, line 341: @Lbqds, does this affect the current code in desktop and mobile wallet for generating the WC storage key for doing several operations like the following?
- getting session topics
- cleaning storage
- cleaning pending requests

https://github.com/alephium/alephium-frontend/blob/next/apps/desktop-wallet/src/features/walletConnect/walletConnectUtils.ts#L35-L37

Lbqds: No, it doesn't affect the wallet. The wallet doesn't use the `WalletConnectProvider` code from the SDK, and since the wallet only needs to handle the Alephium connection without needing to consider other chains, there is no need to use custom storage.

PR #486, opened 2025-01-16
Support for any extension wallet that satisfies the standard

Lbqds:


packages/web3-react/src/utils/injectedProviders.ts
```diff
@@ -106,7 +106,10 @@ export function getInjectedProviderId(provider: AlephiumWindowObject): InjectedP
   if (provider.icon.includes('onekey')) {
     return 'OneKey'
   }
-  return 'Alephium'
+  if (provider.id === 'alephium' && provider.name === 'Alephium') {
+    return 'Alephium'
```
Lbqds, line 110: I will update the doc to specify that new extension wallets are not allowed to use `Alephium` as the wallet name or id.

polarker (approved):
LGTM!

PR #488, opened 2025-01-17
Add Test For Chained Execute Script Transactions

h0ngcha0:


h0ngcha0:
Changed based to https://github.com/alephium/alephium-web3/pull/496, because the test only passes for the full node version later than 3.12.0.

polarker (approved):
LGTM

PR #492, opened 2025-02-14
Ignore non-contract files

Lbqds:


polarker (commented):
LGTM! Please make a patch release for this!

PR #495, opened 2025-02-21
Add getAllContracts method

Lbqds:
This function is used to get the contract list from a third-party project and add it to the current project's contract list, which helps prevent the `Unknown code with code hash` error.


polarker (approved):
LGTM

PR #497, opened 2025-04-28
Fix getChangedSources

Lbqds:
1. When contract `A` extends the abstract contract `B`, if B's code changes, we should be able to recompile contract `A`
2. Add a `--force` flag to the compile command in the CLI


polarker (approved):
LGTM

PR #498, opened 2025-04-30
Improve TestContractParams

Lbqds:
Currently, the parameters we use in unit tests are defined as:
```
export interface TestContractParams<
  F extends Fields = Fields,
  A extends Arguments = Arguments,
  M extends Record<string, Map<Val, Val>> = Record<string, Map<Val, Val>>
> {
  group?: number // default 0
  address?: string
  callerContractAddress?: string
  blockHash?: string
  blockTimeStamp?: number
  txId?: string
  initialFields: F
  initialMaps?: M
  initialAsset?: Asset // default 1 ALPH
  testArgs: A
  existingContracts?: ContractStateWithMaps[] // default no existing contracts
  inputAssets?: InputAsset[] // default no input assets
}
```

For integration tests, we use two types of parameters:
1. Sending transactions via `TxScript`:
```
export interface ExecuteScriptParams<P extends Fields = Fields> {
  initialFields: P
  attoAlphAmount?: Number256
  tokens?: Token[]
  gasAmount?: number
  gasPrice?: Number256
}
```
2. Sending transactions via `transact` method:
```
export interface SignExecuteContractMethodParams<T extends Arguments = Arguments> {
  args: T
  signer: SignerProvider
  attoAlphAmount?: Number256
  tokens?: Token[]
  gasAmount?: number
  gasPrice?: Number256
}
```

Unlike unit tests, integration tests don't need to create contracts, so the parameters are much simpler. Due to this, it's not feasible for the parameters used in unit and integration tests to be the same.

To align them as much as possible, I made the following changes in this PR:

1. Renamed `testArgs` in `TestContractParams` to `args`
2. Renamed `address` in `TestContractParams` to `contractAddress` to avoid ambiguity
3. Added `signer` to `ExecuteScriptParams` for consistency with the parameters used in the `transact` method


packages/web3-wallet/src/hd-wallet.ts
```diff
@@ -98,7 +98,7 @@ export function getHDWalletPath(keyType: KeyType, addressIndex: number): string
   }
   // Being explicit: we always use coinType 1234 no matter the network.
   const coinType = "1234'"
-  const keyTypeNum = (keyType === 'default' || keyType === 'gl-secp256k1') ? 0 : 1
+  const keyTypeNum = keyType === 'default' || keyType === 'gl-secp256k1' ? 0 : 1
```
polarker, line 101: Why do we need to remove the parenthesis?

Lbqds: It was removed by `npm run lint:fix`.

polarker (commented):
LGTM. I will make a new release of full node once the PRs are merged

PR #499, opened 2025-05-02
Danube

polarker:


packages/web3-wallet/src/hd-wallet.test.ts
```diff
@@ -23,18 +23,28 @@ import {
...
 } from './hd-wallet'
 
 describe('HD wallet', () => {
+  const keyTypes: KeyType[] = [
+    'default',
+    'bip340-schnorr',
+    'gl-secp256k1',
+    'gl-secp256r1',
+    'gl-ed25519',
+    'gl-ed25519'
```
polarker, line 45: It should be `gl-webauthn`?

packages/web3/src/api/api-alephium.ts
```diff
@@ -340,18 +368,25 @@ export interface BuildMultisig {
...
+  fromPublicKeyTypes?: string[]
+  fromPublicKeyIndexes?: number[]
   destinations: Destination[]
   /** @format gas */
   gas?: number
   /** @format uint256 */
   gasPrice?: string
+  /** @format group-index */
+  group?: number
+  multiSigType?: MultiSigType
```
polarker, line 380: 1. What's the default value for this?
2. if `fromPublicKeyTypes`, do we treat the address as the lagecy multisig always?

packages/web3/src/api/api-alephium.ts
```diff
@@ -875,6 +958,9 @@ export interface MinerAddressesInfo {
 /** MisbehaviorAction */
 export type MisbehaviorAction = Ban | Unban
 
+/** MultiSigType */
+export type MultiSigType = P2HMPK | P2MPKH
```
polarker, line 962: This is not friendly, we should use enum in Scala.

```
{
  ...,
  multisigType: {
    type: "P2HMPK" | "P2MPKH"
  }
}

vs 

{
  ...,
  multisigType: "P2HMPK" | "P2MPKH"
}
```

packages/web3/src/utils/djb2.ts
```diff
@@ -19,7 +19,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 export default function djb2(bytes: Uint8Array): number {
   let hash = 5381
   for (let i = 0; i < bytes.length; i++) {
-    hash = (hash << 5) + hash + (bytes[`${i}`] & 0xff)
+    hash = ((hash << 5) + hash + (bytes[`${i}`] & 0xff)) | 0
```
polarker, line 22: @Lbqds  Why do we need this change? It may break things

Lbqds: This is because the result returned by `djb2` is inconsistent with the full node. The full node returns an `int32`, but the previous `djb2` implementation returned an `int64`. So we need to use `| 0` to convert it to `int32`, here is a test that illustrates this issue: https://github.com/alephium/alephium-web3/blob/afe6ed61060828b36194b9418a665765bd0971fa/packages/web3/src/utils/djb2.test.ts#L33

packages/web3/src/signer/types.ts
```diff
@@ -119,10 +122,30 @@ export interface SignExecuteScriptTxResult {
...
+
+export type GrouplessSignTxResult<
+  T extends SignExecuteScriptTxResult | SignDeployContractTxResult | SignTransferTxResult
+> = {
+  transferTxs: SignTransferTxResult[]
+} & T
+
+export type SignTxResult<T extends SignExecuteScriptTxResult | SignDeployContractTxResult | SignTransferTxResult> =
+  | GrouplessSignTxResult<T>
+  | T
```
polarker, line 147: Maybe `T` first and then `GrouplessSignTxResult<T>`

packages/web3/src/signer/types.ts
```diff
@@ -34,18 +34,18 @@ export interface Destination {
...
 assertType<Eq<keyof Destination, keyof node.Destination>>
 
-export type KeyType = 'default' | 'bip340-schnorr'
+export type KeyType = 'default' | 'bip340-schnorr' | 'gl-secp256k1' | 'gl-secp256r1' | 'gl-ed25519' | 'gl-webauthn'
 
 export interface Account {
   keyType: KeyType
   address: string
-  group: number
+  group: number // TODO: optional for groupless address
```
polarker, line 42: How about making this optional? If it's groupless, this one is undefined.

packages/web3/src/signer/signer.ts
```diff
@@ -65,9 +67,13 @@ export abstract class SignerProvider {
     }
   }
 
-  abstract signAndSubmitTransferTx(params: SignTransferTxParams): Promise<SignTransferTxResult>
-  abstract signAndSubmitDeployContractTx(params: SignDeployContractTxParams): Promise<SignDeployContractTxResult>
-  abstract signAndSubmitExecuteScriptTx(params: SignExecuteScriptTxParams): Promise<SignExecuteScriptTxResult>
+  abstract signAndSubmitTransferTx(params: SignTransferTxParams): Promise<SignTxResult<SignTransferTxResult>>
```
polarker, line 70: Should we improve the naming of `SignTransferTxResult` too?

packages/web3/src/utils/sign.ts
```diff
@@ -40,7 +40,7 @@ necc.utils.hmacSha256Sync = (key: Uint8Array, ...messages: Uint8Array[]): Uint8A
 export function sign(hash: string, privateKey: string, _keyType?: KeyType): string {
```
polarker, line 40: This function only handles `secp256k1` and `schnorr`
- either we add support for other signatures
- or we limit the KeyType in the function signature

packages/web3/src/api/api-alephium.ts
```diff
@@ -300,12 +288,35 @@ export interface BuildExecuteScriptTx {
...
+  unsignedTx: string
+  /** @format gas */
+  gasAmount: number
+  /** @format uint256 */
+  gasPrice: string
+  /** @format 32-byte-hash */
+  txId: string
+  /** @format address */
+  contractAddress: string
+  transferTxs: BuildSimpleTransferTxResult[]
```
polarker, line 315: As discussed:
1. We'll rename this to `fundingTxs`
2. We'll make this field optional too

packages/web3/src/api/api-alephium.ts
Lbqds: 1. The default value is `P2MPKH`
2. Since we want to reuse the existing multisig API, we introduced the `fromPublicKeyTypes` and `multiSigType` parameters. To use `P2HMPK`, users need to specify both `multiSigType` as `P2HMPK` and `fromPublicKeyTypes`

Correct me if I'm wrong @h0ngcha0

packages/web3/src/utils/sign.ts
```diff
@@ -36,11 +36,18 @@ necc.utils.hmacSha256Sync = (key: Uint8Array, ...messages: Uint8Array[]): Uint8A
   return Uint8Array.from(hash.digest())
 }
 
+function checkKeyType(keyType: KeyType) {
+  if (keyType !== 'default' && keyType !== 'bip340-schnorr' && keyType !== 'gl-secp256k1') {
```
polarker, line 40: Is this condition still correct?

Lbqds: Yes, because we currently only support the secp256k1 and schnorr, perhaps we can also add support for secp256r1 and ed25519 in the future.

packages/web3/src/signer/types.ts
```diff
@@ -34,15 +34,34 @@ export interface Destination {
...
 
+export type Account = GroupedAccount | GrouplessAccount
+
+export function isGroupedAccount(account: Account): account is GroupedAccount {
+  return account.keyType === 'default' || account.keyType === 'bip340-schnorr'
+}
+
+export function isGrouplessAccount(account: Account): account is GrouplessAccount {
+  return account.keyType !== 'default' && account.keyType !== 'bip340-schnorr'
+}
```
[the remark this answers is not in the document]
polarker: Let's do it this way since users/devs might use `xyz:3` even when they want to use a groupless address.

Issue #504, opened 2025-05-21
Update @alephium/web3-wallet to be compatible with more JS environments


nop33:
As it was already done for @alephium/web3, if we want the `@alephium/web3-wallet` package to be usable in non-Node.js enviroments (like browser, React apps, React Native, etc) then we need to polyfill `Buffer`, or replace it with UInt8Array, and replace all deps that use Buffer with others that use UInt8Array.

The following packages use UInt8Array and are the ones we are using in our mobile and desktop wallets:
- `ethereum-cryptography/hdkey`
- `ethereum-cryptography/utils`
- `@metamask/scure-bip39`

nop33:
PR created: https://github.com/alephium/alephium-web3/pull/505

PR #505, opened 2025-05-21
Polyfill buffer in @alephium/web3-wallet

nop33:
- Related to #504


packages/web3-wallet/package.json
```diff
@@ -40,6 +40,7 @@
     "@types/node": "^16.18.23",
     "bip32": "3.1.0",
     "bip39": "3.0.4",
+    "buffer": "^6.0.3",
```
polarker, line 43: How about using a fixed version for this one to mitigate supply chain attacks, since it's a critical dependency for cryptography?

nop33: done

PR #509, opened 2025-05-29
Improve p2pk codec

Lbqds:


packages/web3/src/utils/djb2.ts
```diff
@@ -19,7 +19,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 export default function djb2(bytes: Uint8Array): number {
   let hash = 5381
   for (let i = 0; i < bytes.length; i++) {
-    hash = (hash << 5) + hash + (bytes[`${i}`] & 0xff)
+    hash = ((hash << 5) + hash + (bytes[`${i}`] & 0xff)) | 0
```
Lbqds, line 22: We should ensure that the return value of `djb2` is consistent with the full node, and always stays within the `int32` range.

PR #510, opened 2025-05-29
Add P2HMPK codec

Lbqds:


packages/web3/src/codec/lockup-script-codec.ts
```diff
@@ -37,6 +38,11 @@ export interface P2PK {
   group: number
 }
 
+export interface P2HMPK {
+  hash: PublicKeyHash
```
h0ngcha0, line 42: Since it is not a `PublicKeyHash`, perhaps we can have a new type for it?

PR #513, opened 2025-06-04
Call TxScript with correct group index

Lbqds:


test/contract.test.ts
```diff
@@ -999,4 +999,35 @@ describe('contract', function () {
       ).rejects.toThrow('Test failed due to insufficient funds to cover the dust amount')
     }
   })
+
+  it('should call tx script with the correct group index', async () => {
```
polarker, line 1003: How about deploying to all groups and testing them with a `for` loop?

PR #514, opened 2025-06-11
Support Ralph tuple types

Lbqds:


artifacts/ts/Add.ts
```diff
@@ -41,6 +41,7 @@ import {
   Balances,
   MapValue,
   TokenBalance,
+  TupleTestStruct,
```
polarker, line 44: It seems inefficient that whenever we add a new contract, it's imported into all other contracts' TS files.

PR #518, opened 2025-07-02
Fix signerAddress check in `alph_signAndSubmitChainedTx`

nop33:
- Closes https://github.com/alephium/alephium-web3/issues/519


polarker (approved):
LGTM 👍

PR #520, opened 2025-08-06
Fix `NonExistInput` Error During Chained Txs Submission

h0ngcha0:
When submitting the chained transactions, it is possible that the input doesn't exist yet because a new block isn't produced with the `fromGroup`. 

Here in `signAndSubmitChainedTx`, if we detect that the full node returns the `NonExistInput` error, we try to wait for a block with `fromGroup` to be produced and try again.


h0ngcha0:
Full node handles this correct, infra should be improved to handle chained txs instead. Closing.

PR #523, opened 2025-09-05
Build script tx with correct groupless address

Lbqds:
When calling a contract using a groupless address, for example:

```
MyScript.execute({
  signer,
  initialFields: {
    payer: grouplessAddress
  }
})
```

Since `grouplessAddress` does not have a specified group index, the constructed tx would use the default group index, which may cause the tx to fail. This PR fixes this issue.


polarker (approved):
LGTM 👍

PR #524, opened 2025-09-07
Improve TransactionBuilder to support calling contracts with groupless addresses

Lbqds:
Currently, we have only checked the bytecode and the group in `Script` to support groupless addresses, but the following two situations would still fail:

1. A dApp directly uses `TransactionBuilder` to construct a transaction.
2. A dApp sends a request to the wallet via `WalletConnect`, and the wallet uses `TransactionBuilder` to construct the transaction (which means that once this PR is merged, the wallet will need to update to the latest version of the SDK).

This PR addresses these issues. Additionally, to ensure that the current version of the wallet still works, we need to keep the support in `Script`.

### DIFF
--- a/packages/web3/src/contract/contract.ts
+++ b/packages/web3/src/contract/contract.ts
@@ -42,10 +42,7 @@ import {
   SignExecuteScriptTxParams,
   SignerProvider,
   Address,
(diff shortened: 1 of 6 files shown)


polarker (approved):
LGTM 👍

PR #528, opened 2025-09-25
Disable redirect to deep link on mobile

### LINKED ISSUE #527: Remove deep link to `alephium://`
Now that we have integrated an in-app browser in the mobile wallet, it is not necessary anymore to redirect to `alephium://` deep link when signing. Also, now that there are more wallets in the ecosystem, this prevents them from using our SDK because unless they also update their apps to listen to `alephium://` deep links, the dApps will always redirect to our wallet instead of the other wallets.

nop33:
- Closes #527


polarker (approved):
LGTM 👍

Issue #493, opened 2025-02-17
security issue: Replace non-secure cryptography with secure one


paulmillr:
Instead of increasing test coverage, maybe switch away from elliptic.js; recent elliptic vulnerability leaked private keys from signatures https://github.com/indutny/elliptic/security/advisories/GHSA-vjh7-7g9h-fjfh

polarker:
YES. We will migrate to noble curves, thanks for the wonderful work ❤

Barnabasjohnnie:
I'd love to give this a go.

chidubemOkafor:
Is this issue still available?

Dannynsikak:
i request to be part of the team handling this project , this is an impactful and insightful project and i would love to join in with the devs to fix this issue

Cybermaxi7:
I'd like to tackle the elliptic.js security issue. Having migrated cryptographic libraries in previous blockchain projects, I understand the sensitivity of these changes. I'll implement a switch to noble-curves with minimal disruption to existing code, complete with thorough test coverage. Can deliver a clean, focused PR by next week to eliminate this private key vulnerability.

Issue #446, opened 2024-11-14
Add Native Support for ESM in Packages


h0ngcha0:
Ensure SDK's modules are compatible with ESM, making it easier use and integrate it across the Alephium ecosystem.

Mystic-Nayy:
Hi @h0ngcha0 Can i be assigned to work on this?

ShantelPeters:
I'd like to handle this task.

Supa-mega:
I’d like to help with this.

System625:
Hi, I would love to work on this
I am able to add native support and make sure all modules are compatible for a successful integration

petersdt:
I'd like to take this issue.

PR #544, opened 2026-04-02
Modernize

nop33:
Related to:
- https://github.com/alephium/alephium-web3/issues/446
- https://github.com/alephium/alephium-web3/issues/493
- https://github.com/alephium/alephium-web3/issues/504

### Why

  The `@alephium/web3` package was built with patterns that are increasingly incompatible with modern JavaScript tooling. A developer importing a single function (`isValidAddress`) from the SDK paid nearly the
  same cost as importing the entire library — **742 kB** for a plain website, **928 kB** for a React app. The package shipped a monolithic UMD webpack bundle for browsers, required Node.js polyfills
  (`crypto-browserify`, `stream-browserify`) in every bundler environment, and mutated `BigInt.prototype.toJSON` globally on import — a side effect that prevented bundlers from tree-shaking unused code.

  Getting the SDK to work in React Native / Expo required 12 workarounds including a custom Metro resolver, native crypto modules, Buffer polyfills, and a development build (incompatible with Expo Go).

  This PR modernizes `@alephium/web3` to follow the same patterns used by libraries like [viem](https://github.com/wevm/viem): environment-agnostic, tree-shakeable, and zero side effects.

  ### Results

  | Metric | Before | After | Change |
  |---|---|---|---|
  | Website bundle (vanilla JS, one function + balance fetch) | 742 kB | **48 kB** | **-93%** |
  | Website bundle (gzip) | 204 kB | **14 kB** | **-93%** |
  | Webapp bundle (React, one function + balance fetch) | 928 kB | **238 kB** | **-74%** |
  | Webapp bundle (gzip) | 261 kB | **74 kB** | **-72%** |
  | Direct dependencies | 11 | **6** | -45% |
  | `vite-plugin-node-polyfills` required | Yes | **No** | Removed |
  | ESM output | No | **Yes** | New |
  | Tree-shaking works | No | **Yes** | New |
  | `sideEffects: false` | No | **Yes** | New |
  | UMD browser bundle | 724 kB | **Removed** | -100% |
  | Node polyfill dependencies | 3 | **0** | -100% |
  | `BigInt.prototype.toJSON` global mutation | Yes | **No** | Removed |
  | Minimum Node version | 14 | **20** | Updated |

  ### What changed

  #### Dependencies removed (7)

  - `elliptic` → replaced by `@noble/secp256k1` (already a dependency) + `@noble/curves` for p256/ed25519
  - `bn.js` → was only used by `elliptic`
  - `blakejs` → replaced by `@noble/hashes/blake2b`
  - `crypto-browserify`, `stream-browserify`, `path-browserify` → no longer needed; Node `crypto` import replaced with `globalThis.crypto` (native in Node 20+)
  - `cross-fetch` → native `fetch` available in Node 20+, browsers, and edge runtimes

  #### Dependencies added (2)

  - `@noble/hashes` — isomorphic, audited, zero-dependency hashing (blake2b, sha256, hmac)
  - `@noble/curves` — isomorphic p256 and ed25519 curves

  #### Build system

  - Replaced webpack with **tsup** (`bundle: false`) — each source file is transpiled individually to both CJS (`.js`) and ESM (`.mjs`), preserving the module graph for consumer bundlers to tree-shake
  - Removed the monolithic UMD browser bundle (`alephium-web3.min.js`) — consumers' bundlers now get ESM and perform their own tree-shaking and minification
  - Type declarations generated via `tsc --emitDeclarationOnly` (temporary approach until other workspace packages are migrated to tsup)

  #### Package configuration

  - Proper `exports` field with `"import"` and `"require"` conditions and type declarations
  - `"sideEffects": false` declared — bundlers can safely drop unused modules
  - `"module"` field points to ESM entry for legacy bundler compatibility

  #### Side effect removal

  - Removed `BigInt.prototype.toJSON` global monkey-patch from `src/index.ts`
  - Created `stringify()` utility (inspired by viem) — a drop-in `JSON.stringify` replacement that converts BigInt values to strings via a replacer function
  - Replaced all internal `JSON.stringify` calls with `stringify()`
  - Updated swagger-typescript-api template (`configs/http-client.eta`) and added `scripts/postprocess-schemas.sh` to automatically apply `stringify()` to generated API files after code generation

  #### Other fixes

  - `fs` import in `contract.ts` made dynamic (`await import('fs')`) so it doesn't break browser builds
  - Fixed barrel file re-exports to use `export type` for type-only exports (prevents esbuild from emitting broken ESM references)
  - Updated Node.js engine requirement to >= 20 (Node 18 EOL since April 2025)

  ### What's not in this PR (future work)

  This PR modernizes only `@alephium/web3`. The following packages still need the same treatment:

  - ~~**`@alephium/web3-wallet`** — still uses webpack, `elliptic`, `crypto-browserify`, `buffer`, `bip39` (consider `@scure/bip39`). This is the highest-priority follow-up since it's a direct dependency of most
  consumer apps.~~ done in #545 
  - ~~**`@alephium/walletconnect-provider`** — still uses webpack, `stream-browserify`, `path-browserify`. The `stringify()` sanitization for WalletConnect params is already in place.~~
  - ~~**`@alephium/web3-react`** and **`@alephium/get-extension-wallet`** — use Rollup, relatively modern already.~~ Done in #545 
  - **Sub-path exports** — expose `@alephium/web3/codec`, `@alephium/web3/utils`, etc. as separate entry points for even more granular tree-shaking.
  - **`bignumber.js`** — kept for now (small, zero-dep, used only in `number.ts` for decimal formatting). Can be replaced later.
  - ~~**Type declarations cleanup** — once other workspace packages are migrated to tsup, switch from `tsc --emitDeclarationOnly` to tsup's built-in `dts: true` for cleaner output paths.~~ Not relevant in #545


packages/web3/src/api/utils.ts
```diff
@@ -51,3 +50,16 @@ export function isBalanceEqual(b0: node.Balance, b1: node.Balance): boolean {
     isTokenBalanceEqual(b0.lockedTokenBalances, b1.lockedTokenBalances)
   )
 }
+
+// Drop-in replacement for JSON.stringify that handles BigInt values
+// by converting them to strings. This avoids the need for a global
+// BigInt.prototype.toJSON monkey-patch, enabling "sideEffects": false.
+export const stringify: typeof JSON.stringify = (value: any, replacer?: any, space?: any) =>
```
polarker, line 57: How would handle the cases where a third-party dependency uses JSON.stringify internally?

packages/web3/package.json
```diff
@@ -46,12 +46,12 @@
   },
   "type": "commonjs",
   "dependencies": {
+    "@noble/curves": "^1.9.7",
+    "@noble/hashes": "^1.8.0",
```
polarker: Let's pin the versions for crypto libraries

packages/web3/package.json
```diff
@@ -60,6 +60,8 @@
   },
   "devDependencies": {
     "@babel/eslint-parser": "^7.21.3",
+    "@size-limit/esbuild": "^11.0.0",
+    "@size-limit/file": "^11.0.0",
```
polarker: Why do we suddenly need size-limit as dependencies?

packages/web3/src/utils/sign.ts
```diff
@@ -63,10 +61,11 @@ export function verifySignature(hash: string, publicKey: string, signature: stri
 
   try {
     if (keyType === 'default' || keyType === 'gl-secp256k1') {
-      const key = ec.keyFromPublic(publicKey, 'hex')
-      return key.verify(hash, signatureDecode(ec, signature))
+      const decoded = signatureDecode(signature)
+      const sig = secp.Signature.fromCompact(decoded.r + decoded.s)
```
polarker: It seems we could use secp.Signature.fromCompact(signature) directly, no need to decode now?

packages/web3/package.json
```diff
@@ -46,14 +46,12 @@
   },
   "type": "commonjs",
   "dependencies": {
-    "@noble/curves": "^1.9.7",
-    "@noble/hashes": "^1.8.0",
+    "@noble/curves": "1.6.0",
+    "@noble/hashes": "1.6.1",
```
polarker: It's in generally not a good idea to use such an old version. Are we gonna to upgrade them in your other PRs/commints?

packages/walletconnect/package.json
```diff
@@ -21,11 +21,10 @@
     "walletconnect"
   ],
   "scripts": {
-    "clean": "rm -rf dist",
-    "build:pre": "run-s clean",
+    "clean": "rm -rf node_modules dist",
     "build:cjs": "npx tsc -p tsconfig.cjs.json",
     "build:umd": "webpack",
-    "build": "run-s build:pre build:cjs build:umd",
+    "build": "rm -rf dist && pnpm build:cjs && pnpm build:umd",
```
polarker: Is it a common practice to remove dist everytime we build?

packages/web3/src/contract/contract.ts
```diff
@@ -378,7 +377,8 @@ export class Contract extends Artifact {
     codeHashDebug: string,
     structs: Struct[] = []
   ): Promise<Contract> {
-    const content = await fsPromises.readFile(path)
+    const fs = await import('fs')
```
polarker: Is this a common practice?

packages/web3/package.json
```diff
@@ -37,14 +32,14 @@
     "explorer_backend_version": "3.3.2"
   },
   "scripts": {
-    "build": "rm -rf dist/* && npx tsc --build . && webpack",
+    "clean": "rm -rf node_modules dist",
+    "build": "tsup && tsc --emitDeclarationOnly --declaration --outDir dist/types",
```
polarker: This build script seems not removing the dist folder

package.json
```diff
@@ -19,7 +19,7 @@
     "typescript": "^4.9.5"
   },
   "scripts": {
-    "clean": "rm -rf .parcel-cache && pnpm -r run clean",
+    "clean": "rm -rf node_modules .parcel-cache && pnpm -r run clean",
```
polarker: Is it a common practice to remove node_modules in cleaning? Seems too aggregasive. node_modules likely should not make trouble for Pnpm.

.github/workflows/codeql.yml
```diff
@@ -32,8 +32,8 @@ jobs:
           languages: javascript
 
       - run: |
-          pnpm i
-          npm run build
+          pnpm install --frozen-lockfile
```
polarker, line 35: While you are on it, let's add minimumreleaseage too for the CI: https://pnpm.io/settings#minimumreleaseage

polarker (commented):
Great work, finally we have a proper JS setup for the repo 👍

.github/workflows/codeql.yml
polarker: Could upgrade both .npmrc and pnpm-workspace.yml

Example: https://github.com/ariakit/ariakit/pull/5654/changes#diff-18ae0a0fab29a7db7aded913fd05f30a2c8f6c104fadae86c9d217091709794c

nop33: Done in 538caea7 in:
- https://github.com/alephium/alephium-web3/pull/547

package.json
nop33: It was useful for me for development, especially when dealing with local packages. I can revert it if required. Or better, I can split this script in 2:
```
clean
clean:all
```

packages/walletconnect/package.json
nop33: Yes, for tsc-based builds. tsc doesn't remove stale output files from previous builds. If a source file is renamed or deleted, the old compiled file remains in dist/ and can cause runtime bugs.

packages/web3/package.json
nop33: Good observation. This has been updated in a follow up PR.

packages/walletconnect/package.json
nop33: Cleaned up in ad0a8807 in:
- #547

package.json
nop33: Cleaned up in ad0a8807 in:
- #547

packages/web3/package.json
nop33: It was used for benchmarking the results before and after the modernization PRs. I was considering keeping it and to integrate it in the CI so we can be warned in case our build files exceed a certain size. But maybe that's too much of an optimization for now. Will remove.

EDIT: Done in 9bcd5279

packages/web3/src/contract/contract.ts
nop33: Yes, it's the standard approach for making a library isomorphic when it has Node-only code paths. The alternative is splitting the package into separate Node and browser entry points, which is more complex.

The `fs` usage is only for `Contract.fromArtifactFile()` and `Script.fromArtifactFile()` which are methods that load contract artifacts from the filesystem. These are inherently Node-only operations (since we don't and can't read files from disk in a browser). Making the import dynamic helps to make the import lazy, so that the module is only loaded when the method is actually called, not when the package is imported. Browser consumers who never call `fromArtifactFile()` never trigger the import. The only downside is Metro in React Native, which resolves dynamic imports statically, so RN consumers still need an empty `fs` shim in their Metro config, as you also pointed out here:
- https://github.com/alephium/alephium-web3/pull/547#discussion_r3049119827

However, there's a better approach imo! We could do a dependency injection, similar to what [isomorphic-git does](https://github.com/isomorphic-git/isomorphic-git/blob/846a9f9e7e6ad5e54ee3c38f90c780db5bb178e3/docs/fs.md?plain=1#L8). The 2 functions above could accept a readFile function as a parameter, which would make our SDK fully isomorphic with 0 Node.js builtin refs, and no fs shim needed in RN.

I've implemented this idea in 94400aae in this PR:
-  https://github.com/alephium/alephium-web3/pull/547

Let me know what you think!

polarker: How does viem handle this?

packages/web3/src/api/utils.ts
nop33: If a 3rd-party library calls `JSON.stringify` on an object containing BigInt values, it will throw 
```
TypeError: Do not know how to serialize a BigInt.
```

but that's the correct behavior because it surfaces the problem instead of hiding it in a monkey patch. This is how viem is handling it as well:
- https://github.com/wevm/viem/blob/main/src/utils/stringify.ts

BigInt serialization shouldn't be a hidden global behavior but an explicit decision at the boundary where you convert data for transport or storage. 

From our side, we should of course avoid passing bigint values to 3rd party libraries that are not bigint-aware. If that happens, then we need to convert the data before we pass them.

packages/web3/package.json
nop33: Good call. At the end of the modernization I wanted to upgrade to v2, but I got stuck somewhere and I didn't manage to. I will try that again, and also pin the versions in the last PR.

packages/web3/package.json
nop33: I downgraded then because they use .ts extensions in their declaration files (e.g., `import { BLAKE } from './_blake.ts'`). TypeScript 4.9.5 doesn't support .ts import extensions in .d.ts files, that's a TypeScript 5.x feature. So those versions fail at compile time with 

```
An import path cannot end with a '.ts' extension.
```
    
The sweet spot was v1.6.x, which is CJS compatible and uses .js extensions in declarations, which works with both the TypeScript version and the CJS build setup.

> Are we gonna to upgrade them in your other PRs/commints?

Yes! Now that we have upgraded our TS version, I can upgrade this packages again and the problem should be solved! Good remark 🙏

packages/web3/src/api/utils.ts
polarker: > If that happens, then we need to convert the data before we pass them.

This sounds reasonable, but in practice it's too verbose if the object has 10 nested layers. I believe `BigInt.prototype.toJSON` strikes the best balance in all solutions mentioned here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/BigInt_not_serializable

In general, we should not use such type of injection, but this bigint case is very special.

I am fine with the change since viem does it, but in 2026 still devs reporting such issues on various projects.

packages/web3/src/contract/contract.ts
nop33: Viem doesn't use any Node builtin modules (so it doesn't use `fs`, it doesn't need to read from disk).

polarker: Sounds like a reasonable approach to me! Let's later test this approach with the uniswap-v2-dex and nft prototypes to see if it would affect dev experience much.

packages/web3/package.json
nop33: Upgraded and pined in b3c2c793:
- https://github.com/alephium/alephium-web3/pull/547

packages/web3/package.json
nop33: Upgraded and pined in b3c2c793:
- https://github.com/alephium/alephium-web3/pull/547

The v2/v3 ESM-only packages won't work with our CJS build. To do that we either have to drop CJS output (breaking for require() consumers), or bump minimum Node to 22+ and change the CJS build from `--module commonjs --moduleResolution node` to `--module nodenext --moduleResolution nodenext` (which itself would require additional postprocessing build step, like we have for ESM). Both are significant breaking changes.

Given that EOF of Node 20 is April 2026 (now), maybe we can soon do this upgrade.

polarker: > bump minimum Node to 22+

This should be the approach in the long term. Good that node is evolving to make our life easier

packages/web3/src/utils/sign.ts
nop33: You're right! Fixed in 1602648d:
- https://github.com/alephium/alephium-web3/pull/547

polarker (approved):
LGTM 👍

nop33:
Regarding the failing audit check: We need to eventually update the WalletConnect packages to the latest versions. However, when doing so, tests are failing. There's an issue about it: 
- https://github.com/alephium/alephium-web3/issues/554
See previous discussion at:
- https://github.com/alephium/alephium-web3/pull/552#discussion_r3056385327

PR #545, opened 2026-04-02
Modernize (phase 2)

nop33:
### @alephium/web3-wallet

  Replace all legacy dependencies with modern isomorphic alternatives:
  - `elliptic` → `@noble/secp256k1` (randomPrivateKey)
  - `bip39` → `@scure/bip39` (lighter, no wordlist bloat, Uint8Array-native)
  - `bip32` → `@scure/bip32` (native `@noble` integration, eliminates the entire `noble-wrapper.ts` adapter)
  - Delete `password-crypto.ts` (unused within SDK) 💥 **BREAKING CHANGE**
  - Remove `buffer`, `fs-extra`, `crypto-browserify`, `stream-browserify`
  - Dependencies: 8 → 4

  ### @alephium/walletconnect-provider

  - Remove webpack, UMD bundle, and polyfill devDependencies (`crypto-browserify`, `stream-browserify`, `path-browserify`)
  - No source changes needed — package had no direct Node builtin imports
  - Add proper `exports` with import/require conditions, declare `sideEffects: false`

  ### TypeScript 5.9 upgrade

  - Upgrade from 4.9.5 to 5.9.3 across all 8 packages
  - `moduleResolution: "node"` → `"bundler"` for proper `exports` field resolution
  - `@types/node` 16 → 20
  - Configure ts-jest with CJS override for test compatibility
  - Fix stricter type errors (generic constraints, explicit return types, Uint8Array params)

  ### tsc triple-build (replaces tsup)

  tsup with `bundle: false` had an ESM resolution bug: bare import paths like `"./api/index"` in `.mjs` files resolved to `.js` (CJS) instead of `.mjs`, breaking Vite's dev server silently. Multiple workarounds
  were attempted (explicit paths, `.cjs` extensions, bundled ESM, `outExtension`) but each introduced new issues.

  Switched all three packages to the [same approach as viem](https://github.com/wevm/viem/blob/main/package.json#L8-L10) — three `tsc` passes:
  - `dist/_cjs/` with `{"type":"commonjs"}`
  - `dist/_esm/` with `{"type":"module","sideEffects":false}`
  - `dist/_types/` with declaration files

  All files use `.js` extension; the nested `package.json` `"type"` field controls CJS vs ESM interpretation. This fixes Vite dev server compatibility while maintaining tree-shaking and correct Node.js
  resolution.

  ### Expo app simplified

  Reduced from 12 workarounds to 4:
  - Removed: `buffer`, `events`, `process`, `path-browserify`, `readable-stream`, `react-native-quick-crypto`, `expo-crypto`, `expo-dev-client`, custom Metro resolver
  - Kept: `react-native-get-random-values`, custom entry point, `fs` empty shim (Metro resolves dynamic imports statically), `node-linker=hoisted`
  - **Now works with Expo Go** — no dev build required

  ### Testing infrastructure

  - All four benchmark apps (node-cli, website, webapp, expo) verified working

  ### @noble v2 attempted and reverted

  `@noble/hashes` v2 and `@noble/curves` v2 are ESM-only. The build system handles them fine, but Jest/ts-jest runs in CJS mode and cannot `require()` ESM modules. Upgrading requires migrating from Jest to
  Vitest first.


packages/web3-wallet/package.json
```diff
@@ -31,31 +34,22 @@
...
     "@alephium/web3": "workspace:^",
     "@noble/secp256k1": "1.7.1",
-    "@types/node": "^16.18.23",
-    "bip32": "3.1.0",
-    "bip39": "3.0.4",
-    "buffer": "6.0.3",
-    "elliptic": "6.6.1",
-    "fs-extra": "10.0.1"
+    "@scure/bip32": "^1.2.0",
+    "@scure/bip39": "^1.5.0"
```
polarker: Let's pin the versions for crypto dependencies

packages/cli/package.json
```diff
@@ -28,17 +28,17 @@
...
+    "prettier": "^2.8.7"
   },
   "devDependencies": {
     "@types/jest": "^27.5.2",
-    "@types/node": "^16.18.23",
+    "@types/node": "^20.0.0",
     "@types/prettier": "^2.7.2",
     "ts-node": "^10.9.1",
-    "typescript": "4.7.4"
+    "typescript": "5.9.3"
```
polarker, line 41: What would happen if another project uses an old version of typescript and uses web3 as a dependency? Would that cause issues?

packages/web3-wallet/package.json
nop33: Already pinned in:
- https://github.com/alephium/alephium-web3/pull/547

packages/cli/package.json
nop33: No, it won't cause issues. TypeScript is a devDependency, it's only used at build time in our monorepo. Consumers never use our TypeScript version. They get the compiled .js and .d.ts files from npm, which their own TypeScript version reads. The only thing that matters for consumers is that the .d.ts declarations we emit are compatible with their TypeScript version. TypeScript declarations are backward compatible.

polarker (approved):
LGTM 👍

PR #547, opened 2026-04-07
Modernize/phase 3

nop33:


README.md
```diff
@@ -142,6 +142,35 @@ Runs [publint](https://publint.dev/) and [@arethetypeswrong/cli](https://arethet
...
+module.exports = {}
+```
+
+Add it to your `metro.config.js`:
+```js
+config.resolver.extraNodeModules = {
+  ...config.resolver.extraNodeModules,
+  fs: path.resolve(__dirname, 'shims/fs.js')
+}
+```
```
polarker, line 171: Is it possible to resolve these two issues within the web3 library so that devs won't need to handle them?

pnpm-workspace.yaml
```diff
@@ -1,3 +1,4 @@
 prefer-workspace-packages: true
+minimumReleaseAge: 4320 # 3 days
```
polarker: Let's set 7 days for such a repo. 3 days may be too few around weekend

nop33: Done!

README.md
nop33: Responded in this thread:
- https://github.com/alephium/alephium-web3/pull/544/changes#r3050092718

No need for this anymore!

polarker (approved):
LGTM 👍

PR #550, opened 2026-04-08
Modernize/phase 5 (fix audit warnings)

nop33:


packages/cli/package.json
```diff
@@ -31,7 +31,7 @@
     "crypto-js": "^4.2.0",
     "fetch-retry": "^5.0.6",
     "fs-extra": "^10.1.0",
-    "jest": "^28.1.3",
+    "jest": "^30.3.0",
```
polarker: Why this jest was not replace by vtest?

nop33: This is what I am exploring now! My initial thought was to change as few things as possible for the audit warnings to go away, but it looks like even with the latest jest version they are still there. I am working on replacing it with vitest. I'll move this PR to draft mode.

PR #552, opened 2026-04-08
Fix audit + replace jest with vitest

nop33:


package.json
```diff
@@ -57,7 +57,8 @@
       "elliptic@>=4.0.0 <=6.6.0": ">=6.6.0",
       "elliptic@>=2.0.0 <=6.6.0": ">=6.6.0",
       "elliptic@>=5.2.1 <=6.6.0": ">=6.6.0",
-      "micromatch@<4.0.8": ">=4.0.8"
+      "micromatch@<4.0.8": ">=4.0.8",
+      "lodash@<4.18.0": ">=4.18.0"
     }
```
polarker, line 62: Could you take the chance to clean the overrides list too? Some of them might not be needed any more, e.g. elliptic

nop33: Elliptic is still needed because of WalletConnect. I spent too much time yesterday trying to update the WalletConnect packages (the newest ones don't rely on elliptic anymore) but the tests were failing. I can spend some more time today trying to figure it out.

polarker: Then let's keep it if it's needed. No need to overspend time on this, we could revisit when you have more energy later on )

polarker (approved):
LGTM 👍

package.json
nop33: Deal 🤝 

Issue created:
- https://github.com/alephium/alephium-web3/issues/554

PR #555, opened 2026-04-13
Update WalletConnect to fix audit warnings

nop33:
- Closes https://github.com/alephium/alephium-web3/issues/554

The latest versions of the WalletConnect packages are `2.23.9`. Upgrading to `2.23.9` breaks the tests because the self-hosted relayer (`liuhongchao/relay:hc-dev-3`, a fork of an archived, officially unsupported  WalletConnect repo) doesn't recognize the new JSON-RPC methods (`wc_proposeSession`, `wc_approveSession`) introduced in newer WC versions. WalletConnect explicitly says "self-hosting is not supported.").

Version `2.21.5` is the  latest that works with the existing relay while resolving both audit vulnerabilities (`@stablelib/ed25519` and `elliptic`)

I created a plan for upgrading to `2.23.9` which involves getting rid of the self-hosted relayer:
- https://github.com/alephium/alephium-web3/issues/556

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,28 +46,10 @@
   },
   "pnpm": {
     "overrides": {
-      "semver@<7.5.2": ">=7.5.2",
-      "protobufjs@>=6.10.0 <7.2.4": ">=7.2.4",
-      "@babel/traverse": ">=7.23.2",
-      "browserify-sign": ">=4.2.2",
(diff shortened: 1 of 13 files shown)


h0ngcha0 (approved):
LGTM 👍

PR #563, opened 2026-07-08
Websockets

tdroxler:
CI will fail, because we need first to merge and release websockets on full node.


packages/web3/src/ws/websocket-client.ts
```diff
@@ -0,0 +1,285 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+import WebSocket from 'ws'
+import { EventEmitter } from 'eventemitter3'
+
+import type {
```
nop33, line 21: A `pnpm lint:fix` from root should fix these errors. I included a fix in my PR: c2258a8b8407d9acc954865f0c2672fa07f2083a

packages/web3/src/ws/websocket-client.ts
```diff
@@ -0,0 +1,285 @@
...
+
+The library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+import WebSocket from 'ws'
```
nop33, line 18: `ws` is a Node-only package. `@alephium/web3` was made isomorphic in the series of PRs starting from #544. As of now, if a client-side app uses web3 with these changes, it will [always](https://github.com/websockets/ws/blob/master/package.json#L30) [throw the error](https://github.com/websockets/ws/blob/master/browser.js):
```
ws does not work in the browser. Browser clients must use the native WebSocket object
```
That basically will happen in any environment except Node.js. So all wallets, explorer, and dApps will throw.

An initial mitigation (but doesn't solve the problem completely) is to not use a barrel export, and instead use a subpath export in package.json. That way, only when someone imports from `@alephium/web3/ws` then the Websocket code is loaded. But if someone imports this in any JS env other than Node.js, it will throw again, which again, brakes the isomorphism of the package. Subpath exports is generally a good idea, and also one of my proposed future works on the #544 PR.

A complete fix would be upgrading our Node.js floor from v20 (it went [EOL in April 2026](https://nodejs.org/en/about/previous-releases)) to v22. If we increase `engines.node` to `>=22.4`, we can drop `ws` and `@types/ws` deps, because [Node 22.4 ships](https://nodejs.org/docs/latest-v22.x/api/globals.html#websocket) a (stable, no longer experimental) global `WebSocket`. Browsers as well as RN have one too. It requires some rewrite, however (`this.ws.on('open'|'message'|...)` becomes `addEventListener`, etc).

packages/web3/src/ws/websocket-client.ts
```diff
@@ -0,0 +1,285 @@
...
+  constructor(url: string, options: WebSocketClientOptions = {}) {
+    super()
+    this.requestId = 0
+    this.isConnected = false
+    this.notifications = []
+    this.subscriptionKinds = new Map()
+
+    const headers = options.apiKey ? { 'X-API-KEY': options.apiKey } : undefined
+    this.ws =
+      headers === undefined ? new WebSocket(url) : new WebSocket(url, undefined, { headers })
```
nop33, line 127: Browsers can't send custom headers: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket

This only works in Node.js and in RN (But in RN with [different signature](https://github.com/react/react-native/blob/main/packages/react-native/Libraries/WebSocket/WebSocket.js#L148))

If API-key-protected nodes are meant to be reachable from a browser, the node needs to accept the key via query string or subprotocol. Maybe it can be added in the node PR?

| Environment | Global WebSocket | Custom headers | Signature |
| --- | --- | --- | --- |
| **Browser** | ✅ | ❌ | `(url, protocols)` [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket) |
| **React Native** | ✅ | ✅ | `(url, protocols, { headers })` [RN source](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/WebSocket/WebSocket.js) |
| **Node ≥ 22.4** | ✅ | ✅ | `(url, { headers })` [undici](https://github.com/nodejs/undici/blob/main/docs/docs/api/WebSocket.md)|

packages/web3/src/ws/websocket-client.ts
```diff
@@ -0,0 +1,285 @@
...
+    this.ws.on('open', () => {
+      this.isConnected = true
+      this.emit('connected')
+    })
+
+    this.ws.on('message', (data: WebSocket.Data) => {
+      try {
+        const message: unknown = JSON.parse(data.toString())
+        if (isWsSubscriptionMessage(message)) {
+          this.notifications.push(message)
```
nop33, line 138: Seems like a memory leak. This never gets drained. Also, see below my comments on `onNotification` and `onTypedNotification`

packages/web3/src/ws/websocket-client.ts
```diff
@@ -0,0 +1,285 @@
...
+    this.on('connected', callback)
+  }
+
+  public onNotification(callback: NotificationListener): void {
+    for (const notification of this.notifications) {
+      callback(notification.params)
+    }
+
+    this.on('notification', callback)
+  }
```
nop33, line 239: Since this iterates over the whole history of notifications, if a listeners subscribes 1h after the process has started, they will get a synchronous dump of 1h worth of notifications before starting receiving new ones. Is this intended? Same goes for `onTypedNotification`.

packages/web3/src/ws/websocket-client.ts
```diff
@@ -0,0 +1,285 @@
...
+      this.once(`response_${id}`, (response: WsResponse) => {
+        if ('result' in response) {
+          resolve(response.result as string | boolean)
+          return
+        }
+
+        reject(response.error)
+      })
+
+      this.ws.send(JSON.stringify(request))
```
nop33, line 194: Shouldn't this first check if the client is open? It will throw if it's closed.

packages/web3/src/ws/websocket-client.ts
```diff
@@ -0,0 +1,285 @@
...
+  public onTxNotification(callback: SubscriptionListener<TransactionTemplate>): void {
+    this.onTypedNotification('tx', callback)
+  }
+
+  public onContractEventNotification(callback: SubscriptionListener<ContractEventByBlockHash>): void {
+    this.onTypedNotification('contract', callback)
+  }
+
+  public disconnect(): void {
+    this.removeAllListeners()
```
nop33, line 254: This nukes the consumer's own `on('disconnected')` handler, so they never hear about the disconnect they asked for. I'd remove this line. It's the responsibility of the consumer to remove their listeners if they want.

packages/web3/src/ws/websocket-client.ts
```diff
@@ -0,0 +1,285 @@
...
+        this.emit('error', new Error('Unsupported websocket message'))
+      } catch (error) {
+        this.emit('error', error)
+      }
+    })
+
+    this.ws.on('close', () => {
+      this.isConnected = false
+      this.emit('disconnected')
+    })
```
nop33, line 162: To my understanding, the `close` event gets triggered in 2 cases:
1. The consumer requested it by calling the public `disconnect()` method.
2. The server dropped the connection for whatever reason (server restart, network issues, ...).

In the second case, shouldn't we trigger a reconnection attempt? If we don't, we consumer (Powfi backend, explorer backend) just stop receiving data. Even if we would trigger reconnect on the consumer side, nothing new will be delivered, since  the node forgets what we asked for when the socket dies.

<details><summary>Here's an example</summary>

```ts
  export class WebSocketClient extends EventEmitter {
    private ws: WebSocket
    private url: string
    private isClosing: boolean
    private reconnectAttempts: number
    private subscriptions: Map<string, { kind: WsSubscriptionKind; params: WsSubscriptionParams }>

    constructor(url: string, options: WebSocketClientOptions = {}) {
      super()
      this.url = url
      this.isClosing = false
      this.reconnectAttempts = 0
      this.subscriptions = new Map()
      this.ws = this.connect()
    }

    private connect(): WebSocket {
      const ws = new WebSocket(this.url)

      ws.on('open', () => {
        this.isConnected = true
        const isReconnect = this.reconnectAttempts > 0
        this.reconnectAttempts = 0
        if (isReconnect) {
          this.resubscribeAll()
        } else {
          this.emit('connected')
        }
      })

      ws.on('message', (data) => { /* unchanged */ })
      ws.on('error', (error) => this.emit('error', error))

      ws.on('close', () => {
        this.isConnected = false
        this.emit('disconnected')

        if (this.isClosing) return        // <-- consumer asked for this. stop here.

        this.scheduleReconnect()          // <-- the network did this to us. come back.
      })

      return ws
    }

    private scheduleReconnect(): void {
      const attempt = this.reconnectAttempts++
      const base = Math.min(1000 * 2 ** attempt, 30_000)
      const delay = base * (0.5 + Math.random() / 2)

      setTimeout(() => {
        this.ws = this.connect()
      }, delay)
    }

    private async resubscribeAll(): Promise<void> {
      const active = [...this.subscriptions.values()]
      this.subscriptions.clear()

      for (const { kind, params } of active) {
        const id = await this.subscribe('subscribe', params)
        this.subscriptions.set(id, { kind, params })
      }

      this.emit('reconnected')
    }

    public disconnect(): void {
      this.isClosing = true               // <-- the whole job of disconnect()
      this.ws.close()
    }
  }
```
</details>

I also included an example implementation in my PR (thanks Claude): 2079d64807ebfd8448b0b1d53eb19d3468f61758

packages/web3/src/ws/websocket-client.ts
```diff
@@ -0,0 +1,285 @@
...
+    this.onTypedNotification('tx', callback)
+  }
+
+  public onContractEventNotification(callback: SubscriptionListener<ContractEventByBlockHash>): void {
+    this.onTypedNotification('contract', callback)
+  }
+
+  public disconnect(): void {
+    this.removeAllListeners()
+    this.ws.removeAllListeners()
```
nop33, line 255: This strips the socket's close handler before we call `close()`. `isOpen()` returns true forever after `disconnect()`. I'd also delete this line. I've added a test on this in my PR: 1d1a89234a8a7c56dff8b730978f117d0fa30a24

Also, see my other comment regarding reconnecting.

packages/web3/src/ws/websocket-client.ts
```diff
@@ -0,0 +1,285 @@
...
+      this.isConnected = true
+      this.emit('connected')
+    })
+
+    this.ws.on('message', (data: WebSocket.Data) => {
+      try {
+        const message: unknown = JSON.parse(data.toString())
+        if (isWsSubscriptionMessage(message)) {
+          this.notifications.push(message)
+          const kind = this.subscriptionKinds.get(message.params.subscription)
```
nop33, line 139: I noticed that the node tags every notification with a `type` discriminator that we're not using. We can use it for routing instead of `subscriptionKinds`. That will solve a race condition I reproduced: 1d1a89234a8a7c56dff8b730978f117d0fa30a24

packages/web3/src/fixtures/self-clique.json
```diff
@@ -7,7 +7,6 @@
       {
         "address": "127.0.0.1",
         "restPort": 12973,
-        "wsPort": 11973,
```
nop33, line 10: This looks like it's based on `wsPort` being removed from the API, but the ndoe PR removed it from 
`PeerInfo`, not from `PeerAddress`, which is what self-clique actually returns. On a node built from 
your branch:

```
"nodes": [{"address":"127.0.0.1","restPort":22973,"wsPort":22973,"minerApiPort":20973}]
```

The node's own OpenAPI spec still lists `wsPort` as required on `PeerAddress`, and our regenerated 
`api-alephium.ts:1040` still has `wsPort: number`. So the fixture no longer matches either the type or 
reality.

nop33: Also, this fixture is not used anywhere. Maybe delete?

packages/web3/scripts/ws-listener.ts
```diff
@@ -0,0 +1,127 @@
+import { WebSocketClient } from '../src/ws'
+
+type ListenMode = 'all' | 'block' | 'tx' | 'contract'
+
+const wsUrl = process.env.WS_URL ?? 'ws://127.0.0.1:21973/ws'
```
nop33, line 5: Is this the right port? Tests use `22973`

packages/web3/src/ws/websocket-client.ts
```diff
@@ -0,0 +1,285 @@
...
+    this.subscriptionKinds.clear()
+    this.ws.close()
+  }
+
+  private async subscribeAndRegisterKind(
+    kind: WsSubscriptionKind,
+    params: WsSubscriptionParams
+  ): Promise<string> {
+    const subscriptionId = (await this.subscribe('subscribe', params)) as string
+    this.subscriptionKinds.set(subscriptionId, kind)
```
nop33, line 265: The node registers the subscription and starts sending before we process the response. If a notification lands in that window it'll hit `subscriptionKinds.get(id) === undefined`. A proposed fix is to use the (currently unused `params.type`): 692dd6ee9bc5366204f11230bb954fc824e8d1e7

packages/web3/src/index.ts
```diff
@@ -22,6 +22,7 @@ export * from './signer'
 export * from './utils'
 export * from './transaction'
 export * from './token'
+export * from './ws'
```
nop33, line 25: this is the barrel export I was referring to that it should probably be deleted, see my comment in the beginning.

nop33 (commented):
Nice one! And thanks for the README, it helps.

- A blocker on my side is that the web3 package becomes not isomorphic again. See below.

Some questions:
1. Who is the intended consumer of the TS client? Is it Powfi backend (Node.js env) and explorer backend (Scala, usable only through the npm scripts you wrote)? Or also dApps and wallets? The `@alephium/web3` is currently used by all of these, so we shouldn't break compatibility

I made a PR with some tests and some fixes:
- https://github.com/alephium/alephium-web3/pull/564

packages/web3/src/ws/websocket-client.ts
tdroxler: as discussed with Cheng, we will first upgrade to node 22

